PR #4, opened 2024-10-16
fix: covid academic years off

skvorekn:
## changes

I believe the default COVID academic years are off. Check my thinking here, but Spring 2020 is actually the academic year 2019-2020, and so on.


bdewilde (approved):
excellent catch, my bad 🤦‍♂️

bdewilde:
even offset by a year, the cumnum covid terms was still a predictive feature in my prelim hacc model -- presumably because there's overlap with the _actual_ covid terms. very curious to see how the proper alignment affects the feature's predictive power.

PR #5, opened 2024-10-17
Improve CI workflows

bdewilde:
## changes

- uses a single (non-configurable) version for `uv` in setup action, following [official recommendations](https://docs.astral.sh/uv/guides/integration/github/)
- splits type-check job into its own workflow, no need to double up with tests
- adds a pre-release workflow, ported from other private repo, which checks for a changelog entry
    - adds an initial changelog file to match
- adds a (work-in-progress) publish workflow, for publishing new releases to pypi

## context

Trying to button down package releases, and extend what's been ported from the other repo.

## questions

- Any concerns with automatically publishing a new release when a release is published here on the GitHub repo?
- Who do I talk to about setting this repo up with PyPI tokens for DataKind? Or, alternatively, using the new [trusted publisher](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing) functionality?
- Are there any other checks to do for the pre-release workflow? Like... if possible, should I check that the changelog has an entry for the package's current version? Not sure how strict we want to be here.


.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v45
+        with:
+          files: |
+            CHANGELOG.md
+      - name: Ensure changelog updated
+        if: steps.changed-files.outputs.any_changed == 'false'
+        run: |
+          echo "CHANGELOG.md file must be updated with release notes"
```
skvorekn, line 23: I think this is enough for now!

.github/workflows/test.yml
```diff
@@ -1,34 +1,24 @@
 name: test
 
-on: pull_request
+on:
+  pull_request:  # any pull request
+  schedule:  # run weekly
+    - cron: "0 12 * * 0"
```
skvorekn, line 6: what is the intuition behind running weekly, as well as on PRs? just forward-looking?

CHANGELOG.md
skvorekn, line 1: 🏅

skvorekn (approved):

.github/workflows/test.yml
bdewilde: It's an easy, convenient way to ensure code doesn't break while you're not actively looking at it, e.g. because dependent packages make breaking changes in a patch update.

bdewilde:
> - Any concerns with automatically publishing a new release when a release is published here on the GitHub repo?
> - Who do I talk to about setting this repo up with PyPI tokens for DataKind? Or, alternatively, using the new [trusted publisher](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing) functionality?

Currently, the "release" workflow doesn't actually work, which isn't a blocker rn but is something we need to figure out sooner rather than later. Do you have any thoughts re: these two questions?

skvorekn:
> > * Any concerns with automatically publishing a new release when a release is published here on the GitHub repo?
> > * Who do I talk to about setting this repo up with PyPI tokens for DataKind? Or, alternatively, using the new [trusted publisher](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing) functionality?
> 
> Currently, the "release" workflow doesn't actually work, which isn't a blocker rn but is something we need to figure out sooner rather than later. Do you have any thoughts re: these two questions?

I'm not sure about the PyPI tokens - probably Emma or Larry?

I think an automatic publish makes sense, but I would chat with the rest of the team on a practice here - release iterations will probably be after my time :)

CHANGELOG.md
Eyong20: 🏅

PR #7, opened 2024-10-17
feat: add feature to flag if key courses taken

skvorekn:
## changes

Added the following features:
- num_courses for particular course_ids
- num_courses_grade_DFW
- difference in GPA year 1 - term 1
- change the cohort fixture to have the same IDs as the course fixture, for testing

## context

- Schools often have a sense for which of their courses students get stuck on & hypothesize that they might impact graduation
- DFW is a common aggregation schools are interested in

## questions

As it turns out, installing the local package in Databricks keeps causing my cluster to terminate "unexpectedly" (although at this point it's expected because it keeps happening 😆 ) @bdewilde any suggestions for being able to use the package by...Monday? If you're planning to push the package to PyPI over the weekend, I'll wait for that, but otherwise, I'll probably start trying some other [hacky](https://docs.databricks.com/en/libraries/index.html) options

PR #11, opened 2024-10-22
Rename gpa diff func/feature

bdewilde:
## changes

- renames feature/func, `diff_gpa_year_1_to_term_1` => `diff_gpa_term_1_to_year_1`
- autoformatting applied on save... i wonder why the lint check didn't trigger on that before!

## context

Prior work: PR #7 

As discussed here: https://github.com/datakind/student-success-tool/pull/7#discussion_r1809670692


skvorekn (approved):
I debated the name for this - was trying to decide which was more clear for indicating the order of subtraction! Sounds good.

bdewilde:
> I debated the name for this - was trying to decide which was more clear for indicating the order of subtraction! Sounds good.

Yeah, the underlying calculation had me wondering if this was the right way to phrase it -- sounds like we're both on the same page :D Eventually I concluded that the human way of describing the operation is simply in tension, and that's okay. Better to have it read intuitively for counselors, in the end!

PR #12, opened 2024-10-22
Improve handling for term/year features

bdewilde:
## changes

- removed hacky "num terms checkin" logic from student-term dataset functionality, and replaced it with a more general solution: a "cleanup" function applied to labeled data that dynamically nulls out values coming from years in the future (after the student-term used for prediction) to prevent data leakage
- as a result, included a few extra features that were previously dropped, including the fraction of credits earned year-by-year
- in order for new logic to work, renamed a few features to have a consistent "*_year_X" naming scheme
- also, added a parser to the base pdp cohort dataset schema that nullifies 0 values for "first year to *" fields, which are now included as features
- finally, fixed a bug in a couple CIP-based features that resulted in them being interpreted as numeric values rather than categorical (**note:** this is a minor breaking change)

## context

Still struggling to train a model with performance I'd consider sufficient. Truly grasping at straws. Got any more features to try? I'm all ears!

Also, re: schema change: https://datakind.slack.com/archives/C07KBDVCFUZ/p1729642107713779

## questions

- The logic for nullifying "future" values is a bit tricky. The corresponding unit test makes me more confident it's doing what I want, but please look carefully at this and convince yourself: Is this logic correct?

### DIFF
--- a/src/student_success_tool/analysis/pdp/dataops.py
+++ b/src/student_success_tool/analysis/pdp/dataops.py
@@ -16,7 +16,7 @@ def make_student_term_dataset(
     min_passing_grade: str = constants.DEFAULT_MIN_PASSING_GRADE,
     course_level_pattern: str = constants.DEFAULT_COURSE_LEVEL_PATTERN,
     peak_covid_terms: set[tuple[str, str]] = constants.DEFAULT_PEAK_COVID_TERMS,
-    key_course_subject_areas: t.Optional[list[int]] = None,
+    key_course_subject_areas: t.Optional[list[str]] = None,
     key_course_ids: t.Optional[list[str]] = None,
 ) -> pd.DataFrame:
     """
--- a/src/student_success_tool/analysis/pdp/dataops.py
+++ b/src/student_success_tool/analysis/pdp/dataops.py
@@ -174,9 +174,9 @@ def clean_up_labeled_dataset_cols_and_vals(df: pd.DataFrame) -> pd.DataFrame:
                 "institution_id",
                 "term_id",
                 "academic_year",
-                "academic_term",
+                # "academic_term",  # keeping this to see if useful
                 "cohort",
-                "cohort_term",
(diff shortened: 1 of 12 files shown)


src/student_success_tool/analysis/pdp/dataops.py
```diff
@@ -198,6 +161,84 @@ def standardize_course_dataset(df: pd.DataFrame) -> pd.DataFrame:
...
+    """
+    return (
+        df.pipe(
+            drop_columns_safely,
+            cols=[
+                # metadata
+                "institution_id",
+                "term_id",
+                "academic_year",
+                "academic_term",
```
skvorekn, line 177: Adding some hypotheses in favor of keeping some of these variables, but I don't feel strongly!

Hypothesis around seasonal patterns: students enrolled in the summer look different than other terms

src/student_success_tool/analysis/pdp/dataops.py
```diff
@@ -198,6 +161,84 @@ def standardize_course_dataset(df: pd.DataFrame) -> pd.DataFrame:
...
+                "cohort",
+                "cohort_term",
```
skvorekn, line 179: Hypothesis: students that first enroll in the Fall are different than students that first enroll in the Spring/Summer

skvorekn (approved):
I like it! Pending our assumptions about the PDP column definitions are correct, I feel good about this. I like that you did >= rather than > just to be safe

src/student_success_tool/analysis/pdp/dataops.py
bdewilde: I don't feel strongly, either. Let's err on the side of inclusion -- if the model finds it useful, then so be it.

src/student_success_tool/analysis/pdp/dataops.py
bdewilde: Same here.

PR #14, opened 2024-10-24
Pdp improve grades handling

bdewilde:


skvorekn:
> Ah, shoot. I just realized that this doesn't work with Kelsey's new num_courses_grade_0|1|F|W feature. Thoughts on how to accommodate?

Can you say more about where the logic starts to fail? Is it [here](https://github.com/datakind/student-success-tool/blob/9b7fa3446af4c0f765d3dcc798c7b5b80a2659f1/src/student_success_tool/analysis/pdp/features/student_term.py#L91)?

bdewilde:
> > Ah, shoot. I just realized that this doesn't work with Kelsey's new num_courses_grade_0|1|F|W feature. Thoughts on how to accommodate?
> 
> Hmm, in `sum_val_equal_cols_by_group`, can you allow inputs something like [[[0,1], "F", "W"]](https://github.com/datakind/student-success-tool/blob/9b7fa3446af4c0f765d3dcc798c7b5b80a2659f1/src/student_success_tool/analysis/pdp/features/student_term.py#L99) and check if the value is between [0,1] or equal to "F" or "W"? Am I reading correctly that this is where the logic starts to fail?
> 
> Grade is a weird one because it accepts a variety of types...you could also create a separate function just to handle grade features?

Yup, you clocked it. Your suggested solution would be convenient, but it doesn't work in practice for technical reasons I won't bore you with. Grade _is_ a weird one; I've implemented two solutions; one is slightly more complicated by two orders of magnitude faster, so I'll go with that one :D

bdewilde:
Good news! That latest commit got the num courses failing-or-withdrawal feature working again, and it updated another, existing feature to leverage the same faster code path. This reduced the overall runtime of `pdp.dataops.make_student_term_dataset()` by 2-3x 🥳  Please take another look and let me know what you think.

tests/analysis/pdp/features/test_student_term.py
```diff
@@ -129,36 +129,49 @@ def test_sum_val_equal_cols_by_group(df, grp_cols, agg_col_vals, exp):
...
+    ],
     [
         (
             pd.DataFrame(
                 {
                     "sid": ["123", "123", "123", "456", "456"],
                     "tid": ["22-23 FA", "22-23 FA", "22-23 SP", "22-23 SP", "22-23 SP"],
-                    "course_grade": [4, 3, 1, pd.NA, 4],
-                    "section_course_grade_mean": [3.25, 3.0, 2.5, 3.0, 3.5],
+                    "grade": ["4", "3", "1", pd.NA, "4"],
```
skvorekn, line 146: maybe add some "F" and "W" grades to test those cases?

src/student_success_tool/analysis/pdp/features/student_term.py
```diff
@@ -382,24 +370,62 @@ def sum_val_equal_cols_by_group(
...
-def _rename_sum_by_group_col(col: str) -> str:
-    return f"{constants.NUM_COURSE_FEATURE_COL_PREFIX}_{col}"
+def _course_grade_is_failing_or_withdrawal(
+    df: pd.DataFrame,
+    grade_col: str = "grade",
+    grade_numeric_col: str = "course_grade_numeric",
+) -> pd.Series:
+    return (
+        df[grade_col].isin({"F", "W"})
+        | df[grade_numeric_col].between(0.0, 2.0, inclusive="left")
```
skvorekn, line 422: I think this would include C- per [these grade conversions](https://bigfuture.collegeboard.org/plan-for-college/get-started/how-to-convert-gpa-4.0-scale). Although of course hard to say what PDP grades might look like, want to use 1.7 instead?

skvorekn (changes requested):

src/student_success_tool/analysis/pdp/features/student_term.py
bdewilde: It does not: the bound is left-inclusive, so strictly < 2.0

bdewilde: Wait, what? A C- = 1.7 ??

bdewilde: k I will fix this by passing through the `min_passing_grade` arg, which is how I should've done it originally

bdewilde: thanks for catching this btw! apologies for the scattered comments -- trying to finish modeling/bias assessment and the presentation, and i'm worn out 😮‍💨

skvorekn: All good!! I definitely understand being worn out.

skvorekn (approved):

PR #15, opened 2024-10-27
[PDP] Add term diff features, and improve enrollment year

bdewilde:
## changes

- adds a new family of cumulative features referred to as "term diffs", which compute term-over-term differences for a configurable set of columns; I tried to keep this minimal, and cover three basic indicators: “participation” (num courses taken), “progression” (num credits earned), and “performance” (avg course grade)
- (re-)implements logic to compute start/end dates for terms and cohorts, and uses term+cohort start dates to compute a new "year of enrollment" feature that should align with how PDP computes it in other, similar fields
- adds cleanup logic to nullify column values based on `*_term_X`, in the same way as already implemented for `*year_Y` features
- adds a minimal `types.py` module, in which we can place shared types; starting out with just `TermType`

## context

Continuing to add more info for model training, as possible. All of the year-of-enrollment stuff came out of [this Slack thread](https://datakind.slack.com/archives/C07KBDVCFUZ/p1729964189982969)

Follows PR https://github.com/datakind/student-success-tool/pull/14

## questions

- Is the generic term diff logic correct?
- Is the new term dt / enrollment year logic correct? Would you like me to drop the old enrollment year feature in favor of the v2?


Eyong20 (commented):
reviewed

PR #17, opened 2024-11-02
Upgrade pyproject config/tools

bdewilde:
## changes

Upgrade uv/ruff and leverage new ["dependency groups"](https://peps.python.org/pep-0735/) spec in our python project configuration.

## context

Just keeping our packaging up to date with current Python standards.


kaylawilding (approved):
Simple updates to packaging looks fine

PR #18, opened 2024-11-02
Move [CODENAME] data generation into subpkg?

bdewilde:
## changes

Adds placeholder files for a `generation.CODENAME` subpackage in the sst lib

**TO-DO**
- [ ] refactor and move logic in `synthetic-data/CODENAME/synthetic_data_generator.py` script into `generation.CODENAME` as lib code, in a form similar to what exists for `pdp`
- [ ] generalize `scripts/generate_synthetic_pdp_datasets.py` functionality to be source-agnostic, so we could use the same script to generate synthetic data for either pdp or CODENAME

## context

Currently, synthetic PDP and CODENAME data is generated in similar but different ways, and in two different places in this repo. It would reallly be nice to have these two be consistent and standardized.

## questions

- This isn't an urgent priority, but I do think it's worth doing. Do you feel similarly? Any concerns or objections?

### DIFF
--- a/src/student_success_tool/generation/__init__.py
+++ b/src/student_success_tool/generation/__init__.py
@@ -1,1 +1,1 @@
-from . import pdp
+from . import pdp, zogotech


kaylawilding (approved):
Im not on Zogotech, but I think it makes sense to have a standard/aligned structure for PDP and Zogotech. I think we just need to be careful not to generalize anything without checking it works for both.

bdewilde:
> I think we just need to be careful not to generalize anything without checking it works for both.

Totally agree. Having pdp/zogotech sub-packages living side by side for a bit, with broadly similar structures, will let us better figure out if/how some functionality can be generalized across them.

rachelauryn:
Please rename all references to zogotech to "sample-platform" on our public repo before merging this

bdewilde:
> Please rename all references to zogotech to "sample-platform" on our public repo before merging this

Hey, what's the context for this? [ZogoTech](https://zogotech.com) isn't a secret... but we could always give it a code-name?

nm3224 (approved):

bdewilde:
This PR isn't ready to be merged, or closed. It's a placeholder with a couple notable to-do's that @anzhely never took up. I may revisit in the future, if time permits, but we should leave it open until folks decide otherwise.

PR #19, opened 2024-11-02
[PDP] Refactor base schemas

bdewilde:
## changes

- moves raw cohort+course data schemas out of shared "base" module and into their own so-named modules within the same subpackage
- drops the old base module, and updates imports to match new structure

## context

The original `base.py` module was ported as-is from the other repo; these changes are an attempt to move beyond "legacy" code decisions. :) Specifically, I wanted this updated structure because it aligns with the data generation code in `generation.pdp` -- code consistency is good! -- and because it makes a natural space for adding, say, a `student_term.py` module for validating transformed data before it goes into a model. (That work is out-of-scope here, but on my to-do list.)

## questions

- do you feel strongly about _not_ having duplicate field definitions at the top of these two modules? it's right on the cusp of bugging me; would be easy to create a shared `fields.py` module and import into each schema module, but seemed like a bit more trouble than it's worth.


kaylawilding (approved):
Splitting these out is fine. I think for now its fine to have the duplicate ones and we can backlog putting those in a shared module. I think there are some flexibility issues but I thin they are going to be addressed in PR 28 and beyond.

PR #20, opened 2024-11-02
[PDP] Finish grade feature generalization

bdewilde:
## changes

- adds a course feature that computes a (categorical) grade value by converting numeric grades into classic "letter" grades
- replaces the raw grade with this categorical one where appropriate

## context

This was the last piece needed to fully move away from assuming raw PDP grade values were categoricals. This _happened_ to be the case for HACC, and it also worked albeit differently for MSU, but the PDP spec specifically allows for numeric grades to take continuous values between 0.0 and 4.0. This last piece ensures that we can handle all of these cases with the same featurization code.

## questions

- Any objections to requiring numeric grades (i.e. the `"course_grade_numeric"` col) as inputs here, in addition to the raw grade? I could simplify the inputs and just recompute the numeric grade, since it's a lightweight operation, but I was on the fence about duplicative calculations.

### DIFF
--- a/src/student_success_tool/analysis/pdp/dataops.py
+++ b/src/student_success_tool/analysis/pdp/dataops.py
@@ -193,6 +193,7 @@ def clean_up_labeled_dataset_cols_and_vals(df: pd.DataFrame) -> pd.DataFrame:
                 "term_course_begin_date_min",
                 "term_course_end_date_max",
                 # columns used to derive other features, but not features themselves
+                "grade",
                 "course_ids",
                 "course_subjects",
                 "course_subject_areas",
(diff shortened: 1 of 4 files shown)


kaylawilding (changes requested):
I think this makes sense overall. I believe you are just using this for dealing with like dummies of having an A and or shares of classes with A's and such, but we will still use numeric grade for grade. If that is true, this makes sense to me. 
Only change to note, there is an 'O' as a grade type, should we should include that. I think we should change it to like "OTH" or "OTHER" as you did with audit so its not confused with 0?

bdewilde:
> Only change to note, there is an 'O' as a grade type, should we should include that. I think we should change it to like "OTH" or "OTHER" as you did with audit so its not confused with 0?

Sure, can do. The "A" fix was an actual conflict, but "0"/"O" _looks_ like a conflict.

bdewilde:
Hey @kaylawilding , I've made the change you suggested. Since you're busy today and my schedule is tight bc of the move, I'm going to wield my admin privileges and merge this PR over your "change requested" PR. Please let me know if I've misunderstood your ask and should make other/different changes in a follow-up PR.

PR #22, opened 2024-11-02
[PDP] Add previous term diff cumulative feature(s)

bdewilde:
## changes

Adds "diff-prev-term" features to existing term-diff feature set, which are related to but distinct from the historical term-over-term diffs derived from them.

## context

This sort of feature has been found effective in past projects (@rachelauryn called it out to me a couple weeks ago), I just hadn't had time to implement it. Now, with room to breathe, I realized I was _already computing it_, just tossing it away after using it to derive a similar but different feature.

Follows PR https://github.com/datakind/student-success-tool/pull/21

## questions

- ... Was it really this simple?! Surely I'm forgetting something... 👀


kaylawilding (approved):
I think you are right on adding in the previous term diff features. 

Only thought/question/suggestion, should you not sort the data on student and term before doing the diff transform just to be safe?

bdewilde:
> Only thought/question/suggestion, should you not sort the data on student and term before doing the diff transform just to be safe?

Hey, that's a good question. Since _all_ of these cumulative features are meant to be computed on chronologically sorted data, that's one of the args for the `cumulative.add_features()` function. That actually happens [here](https://github.com/datakind/student-success-tool/blob/develop/src/student_success_tool/analysis/pdp/dataops.py#L72-L76) and then [here](https://github.com/datakind/student-success-tool/blob/develop/src/student_success_tool/analysis/pdp/features/cumulative.py#L18-L19).

This should be safe -- it's not expected that devs are using these lower-level functions. If you're concerns, we can always make the underlying functions "private" (i.e. add a leading underscore to the name), and add docs, so devs know what's up.

PR #28, opened 2024-11-09
[PDP] Generalize base raw data schemas

bdewilde:
## changes

- loosens column checks / nullability requirements that are regularly violated by PDP schools; apparently the official specs are a bit too restrictive
- adds handling to set "unknown"-like values to null for a few more columns, for better consistency and less school-specific sub-classing
- removes inst id from unique config, and instead adds a check for nunique inst ids

## context

We've now seen 3+ PDP schools, so feel reasonably confident that these deviations from the official spec are "generalized"

Follows PR #19


bdewilde:
Hey, I'm confident that these changes are good and a big win over in the school-specific repo. Since I've been using it for USCB and it's just been sitting here for a few days, I'm going to merge it in without review. Post-merge reviews welcome.

PR #30, opened 2024-11-15
[CI] Update uv and enable PyPI publish step

bdewilde:
## changes

- bumps version of `uv` used in GH workflows ([release notes](https://github.com/astral-sh/uv/releases) if curious)
- enables publish package to PyPI GH workflow step, after adding necessary repo secret

## context

This is a prerequisite to publishing a v0.1 of this package to PyPI.

## questions

- How strongly do folks feel about publishing to "test PyPI" before doing an official release? It's a helpful way to catch errors before you make a very public mistake, but not strictly necessary.

### DIFF
--- a/.github/actions/setup-python-env/action.yml
+++ b/.github/actions/setup-python-env/action.yml
@@ -15,7 +15,7 @@ runs:
     - name: Install uv
       uses: astral-sh/setup-uv@v3
       with:
-        version: "0.5.02"
+        version: "0.5.2"
         enable-cache: true
         cache-dependency-glob: "uv.lock"
     - name: Install project and dependencies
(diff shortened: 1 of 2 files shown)


emmaeturner:
I think publishing to "test PyPI" before doing an official release is a great idea! Doesn't hurt to give it a test run before the official release

bdewilde:
> I think publishing to "test PyPI" before doing an official release is a great idea! Doesn't hurt to give it a test run before the official release

Great! :) May I bug you @emmaeturner for a Test PyPI API token, which I'll use in the same way as the (non-test) PyPI API token?

emmaeturner (commented):
I like the idea of publishing to "test PyPI" before doing an official release. Approving these changes!

emmaeturner:
@bdewilde The TEST API token has been added to LastPass and I also Slacked it to you!

PR #31, opened 2024-11-16
[pdp] Add column associations EDA func

bdewilde:
## changes

- adds `compute_pairwise_associations()` function to `eda` module, which is a more generic equivalent to `DataFrame.corr()` that handles nominal-nominal and nominal-numeric column pairs in addition to the usual numeric-numeric
- adds unit tests for said function

## context

I needed a way to compute associations for more data types, since PDP has so many string/categorical variables.

## questions

Nope!

### DIFF
--- a/src/student_success_tool/analysis/pdp/eda.py
+++ b/src/student_success_tool/analysis/pdp/eda.py
@@ -1,6 +1,9 @@
+import itertools
 import typing as t
 
+import numpy as np
 import pandas as pd
+import scipy.stats as ss
 
 from . import utils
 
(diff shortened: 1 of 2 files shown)


kaylawilding (commented):
Cool feature! Maybe since as you said it gets unwieldy, can we drop the columns that are entirely missing or have zero variance prior to creating the graph?

bdewilde:
> Cool feature! Maybe since as you said it gets unwieldy, can we drop the columns that are entirely missing or have zero variance prior to creating the graph?

This is still WIP 🙂 No need to review draft PRs unless specifically requested!

bdewilde:
@kaylawilding I added/formalized functionality to limit which columns are included in the associations matrix. Everything seems to be working as expected! Since you already took a peek and liked what you saw, I'm going to merge this in and keep pushing forward. Hoping to release v0.1 in the next few days 🙏

Eyong20:
great tool

PR #33, opened 2024-11-24
[PDP] Add "retention" target functionality

bdewilde:
## changes

- adds shared functionality for selecting students by next year's course data availability and for selecting the first terms within students' cohort
- adds functionality for computing second-year retention targets, in the same manner as existing targets

## context

A school recently requested that we use this target for SST predictions, so figured I'd properly integrate it into the existing code (and add some unit tests!).

## questions

Nope.


tests/analysis/pdp/targets/test_shared.py
```diff
@@ -9,6 +9,26 @@ def test_df():
     return pd.DataFrame(
         {
             "student_id": ["01", "01", "01", "02", "02", "03", "04", "05"],
+            "cohort_id": [
+                "2020-21 SPRING",
+                "2020-21 SPRING",
+                "2020-21 SPRING",
+                "2021-22 FALL",
+                "2021-22 FALL",
```
Eyong20, line 17: why `2020-21 SPRING` and `2021-22 FALL` repeated?

tests/analysis/pdp/targets/test_shared.py
```diff
@@ -9,6 +9,26 @@ def test_df():
...
+                "2019-20 FALL",
+                "2020-21 SPRING",
+                "2022-23 FALL",
+            ],
+            "term_id": [
+                "2020-21 FALL",
+                "2020-21 SPRING",
+                "2021-22 FALL",
+                "2021-22 FALL",
```
Eyong20, line 26: Same question as above?

tests/analysis/pdp/targets/test_shared.py
bdewilde: These are just test values, and they're repeated because they're for the same student ("01" and "02", respectively). Each row is technically a "student-term", so you'll get repeated values when they're defined at the student level.

Eyong20: ok, understood, great!!

PR #36, opened 2024-11-30
Add student term + labeled data schemas

bdewilde:
## changes

- adds mostly comprehensive schema for pdp student-term datasets
- adds _minimal_ schema for pdp labeled datasets
- upgrade `pandera` package to latest version in lock file

## context

Extending schemas to cover more steps of the pipeline has been on the to-do list for a while; this will give us more confidence about the data going into the modeling step.

## questions

- Any concerns about the not-fully-specified student-term schema here? Making it comprehensive is a pain, and there will be a bit more maintenance burden. It's doable, just... work.


kaylawilding (approved):
Looks good. Im not bothered by the minimal labeled schema. People can add to the schema if needed/can be expanded later. I think its actually easier to add than remove from the schema (if for some reason a feature doesn't make sense in a dataset), so seems fine.

PR #37, opened 2024-11-30
Update codeowners

bdewilde:
## context

We should only include active contributors to this project as codeowners.

## questions

- Does @vishpillai123 want to own anything?


anzhely:
It is my understanding that we will have to change this file again in a week once Daniel and Noreen (the new full time DK employees) get fully onboarded. Also, my understanding that the full time DK folks should own this going forward so Daniel, Noreen and Vish. But Rachel will need to make the call here :)

kaylawilding:
Good with this and what youre saying about the short-term. Lets just add Vish to the pdp and them Im down to approve?

kaylawilding (approved):

PR #38, opened 2024-11-30
Improve labeled dataset cleanup logic

bdewilde:
## changes

refactors how year/term values get masked in cleanup function, such that the input dataframe's dtypes are preserved; this function used to mangle dtypes, which was a pain point in the modeling dataset workflow

## context

Just taking the time to smooth out some rough edges.

## questions

- Are the new functions logical equivalents of the old ones? They _should_ be, but `pandas` is tricksy 🤷‍♂️

### DIFF
--- a/src/student_success_tool/analysis/pdp/dataops.py
+++ b/src/student_success_tool/analysis/pdp/dataops.py
@@ -1,5 +1,6 @@
 import functools as ft
 import logging
+import re
 import typing as t
 
 import pandas as pd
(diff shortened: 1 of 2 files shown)


kaylawilding (approved):
Yes, these seem functionally equivalent in our setting. I think you are good that these are the only features that start with "first_year_to", so none are unintentionally dropped, which would be my main concern with that one. Masking the year columns appear good to me and seem more efficient.

PR #39, opened 2024-12-01
[pdp] Add some new features

bdewilde:
## changes

- adds "term is pre-cohort" feature, and leverages it to improve/streamline pre-cohort term filtering when making a labeled dataset
- adds "num courses in study area" features, both student-term-level and as cumulative fractions

## context

"Pre-cohort" terms came up as a data quality issue for a school, to be filtered out. But it was suggested that we make a feature to note their existence, in case students with pre-cohort terms are meaningfully different (wrt outcome) from those without. This also adds a long-standing feature want, related to the number of courses in students' stated program of study area.

Follows behind PR #38 

## questions

- do these features make sense?
- any other notable features we might want to add?

### DIFF
--- a/src/student_success_tool/analysis/pdp/features/cumulative.py
+++ b/src/student_success_tool/analysis/pdp/features/cumulative.py
@@ -33,6 +33,7 @@ def add_features(
                 ("term_in_peak_covid", "sum"),
                 ("term_is_fall_spring", "sum"),
                 ("term_is_while_student_enrolled_at_other_inst", "sum"),
+                ("term_is_pre_cohort", "sum"),
                 ("course_level_mean", ["mean", "min", "std"]),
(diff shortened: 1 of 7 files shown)


kaylawilding:
Pre-cohort terms is flagging students that took courses before their enrollment term? Do we have an idea of what is happening here? Is it transfers, dual enrollment?

bdewilde:
> Pre-cohort terms is flagging students that took courses before their enrollment term? Do we have an idea of what is happening here? Is it transfers, dual enrollment?

That's correct. For first-time-in-college students, I've only ever seen cases where they had courses in the term immediately prior to enrollment -- seems like something a subset of college students decide to do. A school had a notable subset of transfer students with pre-cohort terms; they're not sure why.

kaylawilding (approved):

PR #40, opened 2024-12-08
Upgrade dependencies/python supported

bdewilde:
## changes

- adds support for PY3.12 in package metadata and CI
- adds/removes/bumps some core package dependencies
- bumps package version, from v0.1.0 to v0.2.0.dev1

## context

Upgraded certain packages to take advantage of newer functionality. Bumped version so semantic versioning and development becomes standard practice for this package. Added PY3.12 since Databricks runtime v16.0 runs it.

## questions

- Any concerns about dropping deps on `textwrap3` and `pillow`? I didn't see them used anywhere...


vishpillai123 (approved):
Looks like a simple/standard upgrade for dependencies. No concerns about pillow or textwrap, I don't see that we're using either anywhere.

PR #41, opened 2024-12-08
[pdp] Extend modeling functionality

bdewilde:
## changes

adds functions (and unit tests) for common modeling-adjacent tasks: splitting datasets into exclusive subsets, and computing sample weights to deal with class imbalance

## context

I've had to do this stuff "custom" for more than one school, but it's the same logic, so should be in the shared lib code.

Follows PR #40

## questions

Are there any other tasks like this to include?

### DIFF
--- a/src/student_success_tool/modeling/utils.py
+++ b/src/student_success_tool/modeling/utils.py
@@ -3,6 +3,7 @@ from collections.abc import Sequence
 
 import numpy as np
 import pandas as pd
+import sklearn.utils
 
 
 def compute_dataset_splits(
@@ -27,7 +28,7 @@ def compute_dataset_splits(
             specify a value for reproducible splits, otherwise each call is unique.
 
     See Also:
-        - :func:`sklearn.model_selection.train_test_split()
+        - :func:`sklearn.model_selection.train_test_split()`
     """
     if len(labels) != len(fracs):
         raise ValueError(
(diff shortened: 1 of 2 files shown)


vishpillai123 (changes requested):
Sample weight function looks to be an extension of `sklearn.utils.class_weight.compute_sample_weight`, so looks great to me.

For the `compute_data_splits()` function, I was curious why using a random generator to create the dataset splits was necessary, instead of also using `sklearn.model_selection.train_test_split`? I have seen train_test_split commonly utilized across academia/industry, so just curious of your choice of implementing the rng library here. Not really requesting changes and I'm definitely open to your thoughts here, just thought to ask. :)

bdewilde:
As the name suggests, `train_test_split()` only splits the dataset into 2 components -- "train" and "test" -- while we want 3 -- the additional "validate" split. It's possible to apply `train_test_split()` twice, but that's slower and clunky. Using the underlying `numpy` functionality directly is faster and accomplishes the same thing. :)

vishpillai123:
> As the name suggests, `train_test_split()` only splits the dataset into 2 components -- "train" and "test" -- while we want 3 -- the additional "validate" split. It's possible to apply `train_test_split()` twice, but that's slower and clunky. Using the underlying `numpy` functionality directly is faster and accomplishes the same thing. :)

Ah, I thought that you could create a validation split and I could have sworn I've used it before, but I'm probably mixing up something else I've used. Ok, makes sense!

vishpillai123 (approved):

PR #43, opened 2024-12-21
[pdp] Map feature columns to human-friendly names

bdewilde:
## changes

- adds `tomli` as optional project dependency (for PY3.10 only)
- adds "assets" directory, and within it, adds a "features table" toml file mapping PDP column names to human-friendly feature names (and, eventually, fuller descriptions)
- adds a utils func to load+parse features tables from disk
- adds feature values to inference topn feature display function (ported from https://github.com/datakind/student-success-intervention/pull/120) and also optional feature name mapping by way of a features table

## context

PDP's internal feature column naming is a bit... technical. Advisors need more friendly naming to help them interpret model outputs correctly.

## questions

- Does this seem like a reasonable implementation?
- How do y'all feel about [TOML](https://toml.io/en/)? Personally I love it, and much prefer it over YAML. :D

### DIFF
--- a/src/student_success_tool/assets/pdp/features_table.toml
+++ b/src/student_success_tool/assets/pdp/features_table.toml
@@ -62,43 +62,35 @@ num_courses_grade_above_section_avg = { name = "number of courses this term with
 cohort_term = { name = "student's cohort (enrollment) term" }
 enrollment_type = { name = "student's enrollment type " }
 enrollment_intensity_first_term = { name = "student's first-term enrollment intensity" }
-math_placement = { name = "TODO" }
-english_placement = { name = "TODO" }
-dual_and_summer_enrollment = { name = "TODO" }
-race = { name = "TODO" }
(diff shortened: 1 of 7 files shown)


src/student_success_tool/assets/pdp/features_table.toml
vishpillai123, line 1: Having human friendly names in our output feature file is a great idea. Definitely onboard with this. I'm also open to .toml file, I don't necessarily have a preference either way.

src/student_success_tool/modeling/inference.py
vishpillai123, line 1: Should we also update the unit test as well under `tests/modeling/test_inference.py`?

Other than that, this looks good to me and similar to the changes in the private repo.

bdewilde: Yeah, good call. I ported those tests over as-is from the private repo; will modernize and extend here.

vishpillai123 (approved):

PR #44, opened 2024-12-22
[pdp] Add project config functionality

bdewilde:
## changes

- adds `pydantic` to build dependencies, used to define generic schemas analogously to `pandera` for datasets
- adds `configs` subpackage, with functionality for loading and validating project configuration from a toml file
- adds (first-draft) config schema for PDP projects
- adds example usage to template nb

## context

This is a next step in standardizing / productionalizing PDP projects: most functionality being driven by a standard, human-friendly config file. We've long had something like this for model training in the private, school-specific repo, but this generalizes / improves upon the concept.

Builds on PR #43, which introduced toml for feature tables.

[Here](https://github.com/datakind/student-success-intervention/blob/c72f86311e7f088e044e37494a7bbae932007fd8/analysis/uscb/config.toml) is an example config file.

## questions

- What do you think of the config schema? Does it make sense / feel intuitive? Is there anything you'd change or that you think is missing? (I have ideas myself, but figured it would be better to share a rough draft to the world before iterating.)

### DIFF
--- a/notebooks/pdp/02-prepare-modeling-dataset-TEMPLATE.py
+++ b/notebooks/pdp/02-prepare-modeling-dataset-TEMPLATE.py
@@ -39,6 +39,7 @@ import pandas as pd
 import seaborn as sb
 from databricks.connect import DatabricksSession
 
+from student_success_tool import configs
 from student_success_tool.analysis import pdp
 
 # COMMAND ----------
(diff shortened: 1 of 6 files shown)


vishpillai123 (approved):
Config implementation will be useful here in terms of ensuring valid inputs to `make_student_term_dataset` and when training/evaluating the model. Config file also looks pretty standard and well-documented.

PR #45, opened 2024-12-27
[pdp] Lowercase all pdp feature names

bdewilde:
## context

This was a bug! Data generated via lib func could have different names compared to the same data read back from a delta table. Folks have run into this in multiple PDP pipelines.

## questions

- is this an acceptable way to handle the problem? bit of a hack, but 🤷

### DIFF
--- a/src/student_success_tool/analysis/pdp/dataops.py
+++ b/src/student_success_tool/analysis/pdp/dataops.py
@@ -5,7 +5,7 @@ import typing as t
 
 import pandas as pd
 
-from . import constants, features, types
+from . import constants, features, types, utils
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/student_success_tool/analysis/pdp/dataops.py
vishpillai123, line 1: standard code, looks good to me

src/student_success_tool/analysis/pdp/utils.py
vishpillai123, line 1: spaces look like typo, but looks fine to me.

vishpillai123 (approved):
overall looks good

src/student_success_tool/analysis/pdp/utils.py
bdewilde: This was just a linting thing -- I'm not sure where those extra whitespace lines came from, but they shouldn't be there.

PR #46, opened 2024-12-29
[pdp] Update template modeling nb to work

bdewilde:
## changes

Updates template train-evaluate-model nb using code in this repo in a way that's compatible with AutoML -- the key was installing `sst` without its dependencies, and instead leveraging those already available in the Databricks Runtime. This is a bit of a hack, but it works, and is closer to what Databricks recommends for AutoML usage.

It also leverages the new project config functionality. The interface is a bit rough right now, but I'll make it better in a next version.

## context

https://datakind.slack.com/archives/C07DWDMN8SW/p1734287236261779

## questions

Does this seem good to you as a first iteration? It should look familiar (similar to the private repo's AutoML nb), if a bit clunky -- I'm going to change and improve functionality later, in separate work.

### DIFF
--- a/notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
+++ b/notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
@@ -165,8 +165,7 @@ if evaluate_model_bias := (training_params.get("split_col") is not None):
         split_col,
     ] + training_params["exclude_cols"]
     df_features = df.drop(columns=non_feature_cols)
-#     train_df[prediction_col] = model.predict(features)
-#     train_df[risk_score_col] = model.predict_proba(features)[:, 1]
+# TODO: figure this out
 # else:
 #     train_df = extract_training_data_from_model(experiment_id)
 #     train_df[prediction_col] = model.predict(train_df)
--- a/notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
+++ b/notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
@@ -23,9 +23,10 @@
 
 # COMMAND ----------
 
-# install dependencies, most of which should come through our 1st-party SST package
+# install (minimal!) extra dependencies not provided by databricks runtime
 # %pip install "student-success-tool==0.1.0"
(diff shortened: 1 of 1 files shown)


notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
```diff
@@ -65,34 +70,30 @@
...
-optional_automl_parameters = {
-    "split_col": "split",
-    "sample_weight_col": "sample_weight",
-    # "pos_label": True,
-    # exclude_frameworks: ["lightgbm", "xgboost"],
-    "timeout_minutes": 5,
-}
-optimization_metric = "log_loss"
+# TODO: prepare school-specific config file in this directory
+config = configs.load_config("./config.toml", schema=configs.PDPProjectConfig)
```
vishpillai123, line 74: should we put a sample .toml file for reference? 

I'm not seeing one pushed in the parent repo, so this errored out for me.

vishpillai123 (changes requested):
The code looks relatively similar to what we have in our private repo, though I'm having some trouble running the notebook without a config.toml file. I was wondering if we're expected to create the config file from scratch?

notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
bdewilde: Well, these templates aren't meant to be runnable as-is -- they're meant to be copied into school-specific contexts where they are then runnable :) The idea of a sample project config seems useful, but I'd consider it out of scope for the changes here!

bdewilde: (I'll see about adding one in a separate PR)

vishpillai123 (approved):

notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
bdewilde: https://github.com/datakind/student-success-tool/pull/50

PR #47, opened 2024-12-30
Fix pdp gen failure from different institution ids

ZakMiller:
There's a cohort schema check that ensures there's a single institution for all the data, but the pdp generation script creates a new institution id per record (so data created using the script will fail the validation). This creates the institution id once and uses it for both cohort and course data.

## changes
- Generates institution id once in the parent script and then passes it down.
- Adds a test that uses the new field to pass the same value down to multiple cohorts/courses (this would fail without this change) for both cohort (which has the check) and course (which doesn't)
- Ranames test_raw_cohort_record to test_raw_course_record in test_raw_course.py (unrelated, looks like a copy paste error)

## context
- Noticed this while running through the templates using the existing test data in the repo. Some more similar changes to follow.

## questions
- Should the same schema check exist for the course schema?
- Where should I create the asana ticket (that's mentioned in CONTRIBUTING.md)?

### DIFF
--- a/scripts/generate_synthetic_pdp_datasets.py
+++ b/scripts/generate_synthetic_pdp_datasets.py
@@ -27,8 +27,7 @@ def main():
     ]
     course_records = [
         FAKER.raw_course_record(
-            cohort_record, normalize_col_names=args.normalize_col_names, institution_id=institution_id
-        )
+            cohort_record, normalize_col_names=args.normalize_col_names)
         for cohort_record in cohort_records
         for _ in range(
             FAKER.randomize_nb_elements(args.avg_num_courses_per_student, min=1)
--- a/scripts/generate_synthetic_pdp_datasets.py
+++ b/scripts/generate_synthetic_pdp_datasets.py
@@ -19,13 +19,15 @@ def main():
     FAKER.add_provider(pdp.raw_cohort.Provider)
     FAKER.add_provider(pdp.raw_course.Provider)
 
+    # institution_id must be the same for all records.
+    institution_id = FAKER.numerify("#####!")
     cohort_records = [
-        FAKER.raw_cohort_record(normalize_col_names=args.normalize_col_names)
+        FAKER.raw_cohort_record(normalize_col_names=args.normalize_col_names, institution_id=institution_id)
         for _ in range(args.num_students)
     ]
     course_records = [
         FAKER.raw_course_record(
-            cohort_record, normalize_col_names=args.normalize_col_names
+            cohort_record, normalize_col_names=args.normalize_col_names, institution_id=institution_id
         )
         for cohort_record in cohort_records
         for _ in range(
--- a/scripts/generate_synthetic_pdp_datasets.py
+++ b/scripts/generate_synthetic_pdp_datasets.py
@@ -22,12 +22,15 @@ def main():
     # institution_id must be the same for all records.
     institution_id = FAKER.numerify("#####!")
     cohort_records = [
-        FAKER.raw_cohort_record(normalize_col_names=args.normalize_col_names, institution_id=institution_id)
+        FAKER.raw_cohort_record(
+            normalize_col_names=args.normalize_col_names, institution_id=institution_id
+        )
         for _ in range(args.num_students)
     ]
     course_records = [
         FAKER.raw_course_record(
-            cohort_record, normalize_col_names=args.normalize_col_names)
+            cohort_record, normalize_col_names=args.normalize_col_names
+        )
         for cohort_record in cohort_records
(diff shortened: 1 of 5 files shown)


src/student_success_tool/generation/pdp/raw_cohort.py
```diff
@@ -12,6 +12,7 @@ def raw_cohort_record(
         min_cohort_yr: int = 2010,
         max_cohort_yr: t.Optional[int] = None,
         normalize_col_names: bool = False,
+        institution_id: int = 12345,
```
bdewilde, line 15:
```suggestion
        institution_id: t.Optional[int] = None,
```

src/student_success_tool/generation/pdp/raw_cohort.py
```diff
@@ -24,7 +25,7 @@ def raw_cohort_record(
         # TODO: handle other cases, e.g. gateway course attempted/completed/grades
         record = {
             "Student GUID": self.student_guid(),
-            "Institution ID": self.institution_id(),
+            "Institution ID": institution_id,
```
bdewilde, line 28:
```suggestion
            "Institution ID": institution_id if institution_id is not None else self.institution_id(),
```

tests/generation/pdp/test_raw_course.py
```diff
@@ -13,10 +13,20 @@
     ["normalize_col_names"],
     [(False,), (True,)],
 )
-def test_raw_cohort_record(normalize_col_names):
+def test_raw_course_record(normalize_col_names):
```
bdewilde, line 16: Looks like the original test name was a copy-paste error, yeah?

src/student_success_tool/generation/pdp/raw_course.py
```diff
@@ -8,7 +8,7 @@
 
 class Provider(BaseProvider):
     def raw_course_record(
-        self, cohort_record: t.Optional[dict] = None, normalize_col_names: bool = False
+        self, cohort_record: t.Optional[dict] = None, normalize_col_names: bool = False, institution_id: int = 12345
```
bdewilde, line 11: I don't feel super strongly about this, but following existing patterns, it seems like we'd want to pull `institution_id` from the `cohort_record` if provided or otherwise generate a random one, rather than passing cohort record && a specified institution id. In this case, I think you'd just leave all the below code as it was, and just drop this extra arg?

bdewilde (commented):
Hey, good catch. Since uniqueness  of `institution_id` only matters when generating multiple records, I think we should make it an optional parameter, so the single-record case still works outside the context of the generate data script. Does that make sense to you? Maybe I'm trying to preserve a use case that doesn't exist in practice, but making it a required arg for both types of records feels a bit heavy.

tests/generation/pdp/test_raw_course.py
ZakMiller: Ya that's what I'm guessing

src/student_success_tool/generation/pdp/raw_course.py
ZakMiller: Oh ya that seems way better. I was missing some context that you added with your comment.

scripts/generate_synthetic_pdp_datasets.py
```diff
@@ -19,14 +19,15 @@ def main():
...
     cohort_records = [
-        FAKER.raw_cohort_record(normalize_col_names=args.normalize_col_names)
+        FAKER.raw_cohort_record(normalize_col_names=args.normalize_col_names, institution_id=institution_id)
         for _ in range(args.num_students)
     ]
     course_records = [
         FAKER.raw_course_record(
-            cohort_record, normalize_col_names=args.normalize_col_names
-        )
+            cohort_record, normalize_col_names=args.normalize_col_names)
```
bdewilde, line 30: Just curious, are you using an autoformatter on this code? This parens _should_ be dangling according to ruff 🤷‍♂️

tests/generation/pdp/test_raw_cohort.py
```diff
@@ -24,3 +24,14 @@ def test_raw_cohort_record(min_cohort_yr, max_cohort_yr, normalize_col_names):
         df_obs = pd.DataFrame([obs])
         obs_valid = RawPDPCohortDataSchema.validate(df_obs, lazy=True)
         assert isinstance(obs_valid, pd.DataFrame)  # => data passed validation
+        print(df_obs)
+
+
+def test_multiple_raw_cohort_records():
```
bdewilde, line 30: nice 👍

bdewilde (approved):

PR #48, opened 2025-01-02
Update inference python file to include shap function

vishpillai123:
SHAP values need parallelization in order to scale for inference. Typically, predictions are generated for hundreds to thousands of students, so it is critical to have this functionality. Spark's repartition does not conserve row order, so the unique identifier needs to be attached every batch.

## changes
- Added "calculate_shap_values" to inference.py. 
- Added respective unit test within tests subdirectory.

## context
- SHAP values are needed for interpretability. Proper scaling is required for inference.

## questions
- No specific questions

### DIFF
--- a/src/student_success_tool/modeling/inference.py
+++ b/src/student_success_tool/modeling/inference.py
@@ -1,9 +1,9 @@
 import typing as t
 from typing import Iterator
+from shap import KernelExplainer
 
 import numpy as np
 import pandas as pd
-from shap import KernelExplainer
 
 def select_top_features_for_display(
     features: pd.DataFrame,
@@ -67,9 +67,13 @@ def select_top_features_for_display(
     return pd.DataFrame(top_features_info)
 
 
-def calculate_shap_values(iterator: Iterator[pd.DataFrame], *, student_id_col: str,
-                model_features: list[str], explainer: KernelExplainer,
-                mode: pd.Series) -> Iterator[pd.DataFrame]:
+def calculate_shap_values(
+    iterator: Iterator[pd.DataFrame], *, 
+    student_id_col: str,
+    model_features: list[str], 
+    explainer: KernelExplainer,
+    mode: pd.Series
+) -> Iterator[pd.DataFrame]:
     """
     SHAP is computationally expensive, so this function enables parallelization,
     by calculating SHAP values over an iterator of DataFrames. Sparks' repartition
--- a/src/student_success_tool/modeling/inference.py
+++ b/src/student_success_tool/modeling/inference.py
@@ -1,8 +1,9 @@
 import typing as t
+from typing import Iterator
 
 import numpy as np
 import pandas as pd
-
+from shap import KernelExplainer
 
 def select_top_features_for_display(
     features: pd.DataFrame,
@@ -64,3 +65,45 @@ def select_top_features_for_display(
                 }
             )
     return pd.DataFrame(top_features_info)
+
+
+def calculate_shap_values(iterator: Iterator[pd.DataFrame], *, student_id_col: str,
+                model_features: list[str], explainer: KernelExplainer,
+                mode: pd.Series) -> Iterator[pd.DataFrame]:
+    """
+    SHAP is computationally expensive, so this function enables parallelization,
+    by calculating SHAP values over an iterator of DataFrames. Sparks' repartition
+    performs a full shuffle (does not preserve row order), so it is critical to 
+    extract the student_id_col prior to creating shap values and then reattach 
+    for our final output.
+
+    Args:
+        iterator (Iterator[pd.DataFrame]): An iterator over Pandas DataFrames.
+        Each DataFrame is a batch of data points.
+        student_id_col (str): The name of the column containing student_id
+        model_features (list[str]): A list of strings representing the names 
+        of the features for our model
+        explainer (shap.KernelExplainer): A KernelExplainer object used to compute
+        shap values from our loaded model.
+        mode (pd.Series): A Series containing values to impute missing values
+
+    Returns:
+        Iterator[pd.DataFrame]: An iterator over Pandas DataFrames. Each DataFrame
+        contains the SHAP values for that partition of data.
+    """
+
+    for pdf in iterator:
+        # Preserve student_id column
+        student_ids_batch = pdf.loc[:, student_id_col]
+
+        # Impute missing values and run shap values using just pdf features
+        pdf_features = pdf[model_features].fillna(mode)
+        shap_values = explainer.shap_values(pdf_features)
+
+        # Create a DataFrame from the SHAP values
+        shap_df = pd.DataFrame(shap_values, columns=model_features)
+
+        # Reattach the student_id column to our SHAP values DataFrame
+        shap_df[student_id_col] = student_ids_batch
+
+        yield shap_df
--- a/src/student_success_tool/modeling/inference.py
+++ b/src/student_success_tool/modeling/inference.py
@@ -1,5 +1,4 @@
 import typing as t
-from typing import Iterator
 from shap import KernelExplainer
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


src/student_success_tool/modeling/inference.py
```diff
@@ -64,3 +65,49 @@ def select_top_features_for_display(
...
+        explainer (shap.KernelExplainer): A KernelExplainer object used to compute
+        shap values from our loaded model.
+        mode (pd.Series): A Series containing values to impute missing values
+
+    Returns:
+        Iterator[pd.DataFrame]: An iterator over Pandas DataFrames. Each DataFrame
+        contains the SHAP values for that partition of data.
+    """
+
+    for pdf in iterator:
```
bdewilde, line 99: What's "pdf" stand for? I think of the file format... Maybe you think "pandas data frame"? If you named the input arg `dfs`, then this would be an easy `for df in dfs:` :)

src/student_success_tool/modeling/inference.py
```diff
@@ -64,3 +65,49 @@ def select_top_features_for_display(
...
+
+def calculate_shap_values(
+    iterator: Iterator[pd.DataFrame], *, 
+    student_id_col: str,
+    model_features: list[str], 
+    explainer: KernelExplainer,
+    mode: pd.Series
+) -> Iterator[pd.DataFrame]:
+    """
+    SHAP is computationally expensive, so this function enables parallelization,
```
bdewilde, line 78: Dumb question: Where is the parallelization happening? To me, it looks like this function is iterating over the dataframes one after the other in a `for` loop.

src/student_success_tool/modeling/inference.py
```diff
@@ -64,3 +65,49 @@ def select_top_features_for_display(
...
+    for pdf in iterator:
+        # Preserve student_id column
+        student_ids_batch = pdf.loc[:, student_id_col]
+
+        # Impute missing values and run shap values using just pdf features
+        pdf_features = pdf[model_features].fillna(mode)
+        shap_values = explainer.shap_values(pdf_features)
+
+        # Create a DataFrame from the SHAP values
+        shap_df = pd.DataFrame(shap_values, columns=model_features)
```
bdewilde, line 108: It's a bit irritating that shap converts dataframes into raw numpy arrays under the hood, so we have to reconstruct the tabular data ://

tests/modeling/test_inference.py
```diff
@@ -115,3 +115,91 @@ def test_select_top_features_for_display(
...
+        ({"student_id": [1, 2], "feature1": [0.1, 0.2], "feature2": [0.4, 0.5]}, (2, 3))
+    ]
+)
+def test_calculate_shap_values_basic(input_data, expected_shape, explainer):
+    df = pd.DataFrame(input_data)
+    student_id_col = 'student_id'
+    model_features = ['feature1', 'feature2']
+    mode = df.mode().iloc[0] 
+    
+    iterator = iter([df])
```
bdewilde, line 152: This is the thing I mentioned before -- doesn't this function work equally well if you pass `[df]` in directly, rather than converting it into an iterator?

tests/modeling/test_inference.py
```diff
@@ -115,3 +115,91 @@ def test_select_top_features_for_display(
...
+    # Check that the result contains the expected number of rows and columns
+    shap_df = result[0]
+    assert shap_df.shape == expected_shape
+    
+    # Ensure that 'student_id' column is present
+    assert student_id_col in shap_df.columns
+    
+    # Ensure that SHAP values are generated and are numeric
+    assert isinstance(shap_df[model_features].iloc[0, 0], (float, np.float64))
+    assert isinstance(shap_df[model_features].iloc[0, 1], (float, np.float64))
```
bdewilde, line 165: There may be handy pandas functionality for this -- maybe `from pandas.api.types import is_numeric_dtype`

bdewilde (changes requested):

src/student_success_tool/modeling/inference.py
```diff
@@ -1,9 +1,10 @@
 import typing as t
+from typing import Iterator
```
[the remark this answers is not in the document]
vishpillai123: Will do.

src/student_success_tool/modeling/inference.py
```diff
@@ -64,3 +65,49 @@ def select_top_features_for_display(
                 }
             )
     return pd.DataFrame(top_features_info)
+
+
+def calculate_shap_values(
+    iterator: Iterator[pd.DataFrame], *,
```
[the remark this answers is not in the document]
vishpillai123: So this is a copy of what we have in the private repo, in which the legacy function we previously called an "Iterator". Perhaps it is more of an iterable. Either way, I can rename the variables descriptively rather than by type as you are suggesting.

src/student_success_tool/modeling/inference.py
```diff
@@ -64,3 +65,49 @@ def select_top_features_for_display(
...
+
+    Args:
+        iterator (Iterator[pd.DataFrame]): An iterator over Pandas DataFrames.
+        Each DataFrame is a batch of data points.
+        student_id_col (str): The name of the column containing student_id
+        model_features (list[str]): A list of strings representing the names 
+        of the features for our model
+        explainer (shap.KernelExplainer): A KernelExplainer object used to compute
+        shap values from our loaded model.
+        mode (pd.Series): A Series containing values to impute missing values
```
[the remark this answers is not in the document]
vishpillai123: Good point - sounds good

src/student_success_tool/modeling/inference.py
```diff
@@ -64,3 +65,49 @@ def select_top_features_for_display(
...
             )
     return pd.DataFrame(top_features_info)
+
+
+def calculate_shap_values(
+    iterator: Iterator[pd.DataFrame], *, 
+    student_id_col: str,
+    model_features: list[str], 
+    explainer: KernelExplainer,
+    mode: pd.Series
```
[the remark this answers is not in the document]
vishpillai123: sg

src/student_success_tool/modeling/inference.py
vishpillai123: yes, `for df in dfs` is more straightforward.

src/student_success_tool/modeling/inference.py
vishpillai123: So, I just copied over our function from the private repo. I was planning on merging this PR and then working on adding the SHAP values notebook, which is where the parallelization is happening using `spark.repartition`. Not a dumb question - You are correct in that parallelization is not happening in this function as it is just an iteration. 

Do you think it's best to create a notebook or put all parallelization in this function? And if we want to create a notebook, I can keep it outside of the PDP template notebooks, or as the 4th template notebook. Curious of your thoughts?

tests/modeling/test_inference.py
vishpillai123: This is used in context of spark's repartition in which I believe that would need an iterator. You are right in this dummy case of our unit test, it doesn't serve a purpose.

src/student_success_tool/modeling/inference.py
bdewilde: I guess I don't understand how `spark.repartition` works? Normally I'd expect an operation to-be-parallelized to have a function that operates on one "chunk" of the iterable, and then some outside framework calls that function in parallel over chunks. How does it work if the iterable is _inside_ the function?

Apologies if I'm being dense, I am properly confused! 😅

bdewilde: Is this what's going on here? https://www.databricks.com/blog/2020/05/20/new-pandas-udfs-and-python-type-hints-in-the-upcoming-release-of-apache-spark-3-0.html

In all the docs, I only see them supporting `Iterator[pd.Series]` inputs, rather than `Iterator[pd.DataFrame]`, so even if this what's being used here, I'm still confused!

bdewilde: Actually maybe this post is the closest analogue? https://www.databricks.com/blog/2022/02/02/scaling-shap-calculations-with-pyspark-and-pandas-udf.html

bdewilde (approved):
I read the full blog post from Databricks that I linked to -- looks like we're mostly just doing what they recommend in there. imo the API for Pandas UDFs in Spark is... weird, and was the cause of my immediate confusion. I believe this should work, so please merge it. That said, I'm going to come around after you and tidy this up in a way that makes it more obvious what's going on. Does that sound okay to you?

bdewilde:
@vishpillai123 I re-ran the CI jobs -- there's a test that's juuuuust slightly flakey -- and they all pass. I'm going to merge this in, since I believe I understand what's going on, so I can refactor/rename the logic for clarity. Please forgive if I'm jumping the gun here!

vishpillai123:
hey no worries. I had to jump onto some other work with inference generation being the priority this month. I know this is still pretty incomplete, but I think we need to add a notebook to provide some clarity. Otherwise, this function is quite confusing to use on its own. I'll start working on that soon pending some other priorities.

PR #49, opened 2025-01-03
Allow True and False through _cast_to_bool_via_int unchanged

ZakMiller:
The checked in sample data in INSTXYZ_STUDENT_SEMESTER_AR_DEIDENTIFIED.csv has "True" and "False" for the retention and persistence columns. The parsing logic assumes those columns will be numbers.

## changes
- Change _cast_to_bool_via_int to use a map replace so it handles True, False, 1, and 0. The switching to an object type first is necessary to allow the boolean values.
- Also added a new overall parsing test case that basically matches what's checked in (I replaced the cohort ids with a consistent cohort id, as that's another validation failure, see https://github.com/datakind/student-success-tool/pull/47

## context
Noticed this while running through the templates using the existing test data in the repo. Some more similar changes to follow.


src/student_success_tool/analysis/pdp/dataio.py
```diff
@@ -182,8 +182,7 @@ def _replace_values_with_null(
 
 
 def _cast_to_bool_via_int(df: pd.DataFrame, *, col: str) -> pd.Series:
-    return df[col].astype("Int8").astype("boolean")
-
+    return df[col].astype('object').replace({"1": True, "0": False, "True": True, "False": False}).astype('boolean')
```
bdewilde, line 185: nit: code style

```suggestion
    return df[col].astype("object").replace({"1": True, "0": False, "True": True, "False": False}).astype("boolean")
```

bdewilde (commented):
I'm not sure I understand the need for this. Are you saying that we sometimes get boolean data passed in as "True" and "False"? Or just that the "test" data is slightly different from what we actually get from PDP, in the sense that we use True/False (read in as strings) whereas PDP uses 1/0?

_Update:_ Okay, coffee has kicked in, I understand what's going on now. Pandas makes it slightly awkward to handle  this -- generally downcasting into "object" dtypes is discouraged afaik. Let me think... What we'd really like is a pandas function equivalent to `to_datetime()` or `to_numeric()`, only `to_boolean()`. They must have this _somewhere_ under the hood, considering the options that `pd.read_csv()` exposes. I wonder if there's some some secret API...

bdewilde (commented):
Alright, this is the only way I could find that wouldn't require downcasting into "object" dtypes:

```python
    return (
        df[col].astype("string")
        .map({"1": True, "0": False, "True": True, "False": False, "true": True, "false": False})
        .astype("boolean")
    )
```

What do you think? Is that an improvement over your approach, or just a detour? :)

ZakMiller:
That suggestion works for me!

bdewilde (approved):
:shipit:

PR #50, opened 2025-01-04
[pdp] Add prelim template project config file

bdewilde:
## changes

- Adds a template config file to the notebooks dir, alongside other templates that rely on the config
- Adds a config test fixture file and adds a test for loading that config

## context

Flagged in another PR — having a config template helps folks to use the template notebooks

## questions

- Any objections to including the config template inside the notebooks dir? Slightly awkward name-wise, but functionality-wise makes sense. We could consider renaming `notebooks` to something like, oh, `projects` or `templates`, if this really bugs you!
- Is anyone using this project config functionality in their school pipelines yet? (Besides me for USCB.) I'm thinking of changing the schema, but don't want to break anyone's pipelines. If the answer is yes, I'll push the schema change into the version after the next release of this package, nbd.

### DIFF
--- a/notebooks/pdp/config-TEMPLATE.toml
+++ b/notebooks/pdp/config-TEMPLATE.toml
@@ -1,0 +1,22 @@
+institution_name = "SCHOOL_NAME"
+
+[prepare_modeling_dataset]
+min_passing_grade = 1.0
+min_num_credits_full_time = 12.0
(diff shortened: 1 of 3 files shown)


nm3224 (commented):
looks great!

bdewilde:
Thanks @nm3224 ! If you think it's "merge-ready", please give it an "approve" review rather than just a "comment" review. I need an approve in order to merge :)

nm3224 (approved):

PR #52, opened 2025-01-10
Fix PDP synthetic generation year format

ZakMiller:
## changes
- Switch from XXXX-XXXX to XXXX-XX because that's the schema
- Refreshed the synthetic data checked in to the repo with the changes
- Test loading the synthetic data checked into the repo (which should prevent the two from deviating in the future)

## context
- Makes the sample generation code consistent with the schema logic

## questions
- n/a


src/student_success_tool/generation/pdp/raw_cohort.py
```diff
@@ -163,7 +163,7 @@ def cohort(self, min_yr: int = 2010, max_yr: t.Optional[int] = None) -> str:
             start_date=date(min_yr, 1, 1), end_date=_end_date
         )
         end_dt = start_dt.replace(year=start_dt.year + 1)
-        return f"{start_dt:%Y}-{end_dt:%Y}"
+        return f"{start_dt:%Y}-{end_dt:%y}"
```
bdewilde, line 166: nice and simple 👍

src/student_success_tool/generation/pdp/raw_course.py
```diff
@@ -150,7 +150,7 @@ def academic_year(self, min_yr: int = 2010, max_yr: t.Optional[int] = None) -> s
             end_date=(date(max_yr, 1, 1) if max_yr else "today"),
         ).replace(day=1)
         end_dt = start_dt.replace(year=start_dt.year + 1)
-        return f"{start_dt:%Y}-{end_dt:%Y}"
+        return f"{start_dt:%Y}-{end_dt:%y}"
```
bdewilde, line 153: 3x logic duplication is right on the cusp of wanting a consolidated function/method. I think it's fine as-is, since we don't expect to ever have, say, 4x duplication?

tests/analysis/pdp/test_dataops.py
```diff
@@ -230,3 +233,35 @@ def test_clean_up_labeled_dataset_cols_and_vals(df, exp):
...
+@pytest.mark.parametrize(
+    ["cohort_file_name", "course_file_name"],
+    [
+        (
+            "INSTXYZ_STUDENT_SEMESTER_AR_DEIDENTIFIED.csv",
+            "INSTXYZ_COURSE_LEVEL_AR_DEID.csv",
+        ),
+    ],
+)
+def test_make_student_term_dataset_against_checked_in_sample(
```
bdewilde, line 247: seems like a good thing to have, thanks for adding

bdewilde (approved):
:shipit: 

I may look into generalizing year-handling to allow for YYYY-YY or YYYY-YYYY formats, but this is the better approach short-term.

bdewilde:
@ZakMiller I'm going to merge this in for you, since I do SST work more on weekends but don't expect the same from you :) Thanks for submitting this fix!

PR #53, opened 2025-01-11
Update inference output file per discussion

bdewilde:
## changes

- changes output data structure from "long" to "wide", features-wise
- adds optional logic to compute "support needed" bool values from probs
- updates unit tests, including a little bit for new float rounding

## context

Cousin PR: https://github.com/datakind/student-success-intervention/pull/127

## questions

This is similar to but different from what you implemented in the other repo. My intention is that this replaces that in the near future. Do you have any concerns about the differences?

### DIFF
--- a/src/student_success_tool/modeling/inference.py
+++ b/src/student_success_tool/modeling/inference.py
@@ -26,7 +26,14 @@ def select_top_features_for_display(
             order as unique_ids, of shape len(unique_ids)
         shap_values: array of arrays of SHAP values, of shape len(unique_ids)
         n_features: number of important features to return
-        needs_support_threshold_prob
+        needs_support_threshold_prob: Minimum probability in [0.0, 1.0] used to compute
+            a boolean "needs support" field added to output records. Values in
+            ``predicted_probabilities`` greater than or equal to this threshold result in
+            a True value, otherwise it's False; if this threshold is set to null,
+            then no "needs support" values are added to the output records.
+            Note that this doesn't have to be the "optimal" decision threshold for
+            the trained model that produced ``predicted_probabilities`` , it can
+            be tailored to a school's preferences and use case.
         features_table: Optional mapping of column to human-friendly feature name/desc,
             loaded via :func:`utils.load_features_table()`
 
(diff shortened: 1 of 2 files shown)


src/student_success_tool/modeling/inference.py
```diff
@@ -54,16 +65,18 @@ def select_top_features_for_display(
...
-                {
-                    "Student ID": unique_id,
-                    "Support Score": predicted_proba,
-                    "Top Indicators": feature_name,
-                    "Indicator Value": feature_value,
-                    "SHAP Value": shap_value,
-                    "Rank": rank,
-                }
+            feature_value = (
+                str(round(feature_value, 2))
```
vishpillai123, line 69: 2 works :)

src/student_success_tool/modeling/inference.py
```diff
@@ -9,8 +10,9 @@ def select_top_features_for_display(
     features: pd.DataFrame,
     unique_ids: pd.Series,
     predicted_probabilities: list[float],
-    shap_values: pd.Series,
+    shap_values: npt.NDArray[np.float64],
     n_features: int = 3,
+    needs_support_threshold_prob: t.Optional[float] = 0.5,
```
vishpillai123, line 15: I like two inputs being consolidated into one, though just want to make sure we have documentation around this. It's somewhat explanatory but I think it's still good to add description below.

vishpillai123 (approved):
hey, everything looks good to me! If we can just add some more documentation around the `needs_support_threshold_prob`.

src/student_success_tool/modeling/inference.py
```diff
@@ -24,6 +26,7 @@ def select_top_features_for_display(
             order as unique_ids, of shape len(unique_ids)
         shap_values: array of arrays of SHAP values, of shape len(unique_ids)
         n_features: number of important features to return
+        needs_support_threshold_prob
```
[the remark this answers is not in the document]
bdewilde: yup, good call. thanks for not letting me cut corners ;)

PR #56, opened 2025-01-23
Improve pdp project config structure + functionality

bdewilde:
## changes

- adds a "v2" pdp project configuration schema, which differs from the previous iteration in structure and scope, plus additional validation and documentation
- adds an example template file for the v2 config
- adds a unit test to check that the schema is working correctly

Out-of-scope: Using this config in nbs. I tried that in another PR, but it got much too big.

## context

This file is meant to record the full set of configuration options for a given school's pipeline in a way that's consolidated, human-readable, and not just hard-coded magic variables spread across multiple notebooks. It's especially useful when parameters must be shared and consistent across nbs; consolidating them here makes that task much easier and less error-prone.

The "v1" config was a WIP, but then a couple schools over in the private repo started using it, and I didn't want to break their shit. So, here's a "v2". In the next minor (not patch) release of this package, I'll finish the transition and make this v2 into the only option. A bit awkward, but only temporary, for the sake of a smooth transition.

## questions

- What do you think of the config's structure and contents?
- Is there additional validation I should be doing, to ensure things make sense?

### DIFF
--- a/notebooks/pdp/config-v2-TEMPLATE.toml
+++ b/notebooks/pdp/config-v2-TEMPLATE.toml
@@ -1,0 +1,52 @@
+institution_id = "INST_ID"
+institution_name = "INST NAME"
+
+student_id_col = "student_guid"
+target_col = "target"
+split_col = "split"
+sample_weight_col = "sample_weight"
+student_group_cols = ["student_age", "race", "ethnicity", "gender", "first_gen"]
+pred_col = "pred"
+pred_prob_col = "pred_prob"
+pos_label = true
+random_state = 12345
+
+[datasets.labeled]
+raw_course = { file_path = "/Volumes/CATALOG/INST_NAME_bronze/INST_NAME_bronze_file_volume/FILE_NAME_COURSE.csv" }
+raw_cohort = { file_path = "/Volumes/CATALOG/INST_NAME_bronze/INST_NAME_bronze_file_volume/FILE_NAME_COHORT.csv" }
+preprocessed = { table_path = "CATALOG.SCHEMA.TABLE_NAME" }
+
+[preprocessing]
+splits = { train = 0.6, test = 0.2, validate = 0.2 }
+sample_class_weight = "balanced"
+
+[preprocessing.features]
+min_passing_grade = 1.0
+min_num_credits_full_time = 12
+# NOTE: single quotes *required* here; it's TOML syntax for literal strings
+course_level_pattern = 'asdf'
+key_course_subject_areas = ["24", "51"]
+key_course_ids = ["ENGL101", "MATH101"]
+
+[preprocessing.target]
+params = { min_num_credits_checkin = 30.0, min_num_credits_target = 60.0 }
+student_criteria = { enrollment_type = "FIRST-TIME", credential_type_sought_year_1 = "Bachelor's Degree" }
+
+[modeling.feature_selection]
+incomplete_threshold = 0.5
+low_variance_threshold = 0.0
+collinear_threshold = 10.0
+
+[modeling.training]
+# exclude_frameworks = ["xgboost", "lightgbm"]
+primary_metric = "log_loss"
+timeout_minutes = 10
+
+[trained_model]
+experiment_id = "EXPERIMENT_ID"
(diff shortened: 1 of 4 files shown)


notebooks/pdp/config-v2-TEMPLATE.toml
```diff
@@ -0,0 +1,52 @@
...
+
+student_id_col = "student_guid"
+target_col = "target"
+split_col = "split"
+sample_weight_col = "sample_weight"
+student_group_cols = ["student_age", "race", "ethnicity", "gender", "first_gen"]
+pred_col = "pred"
+pred_prob_col = "pred_prob"
+pos_label = true
+random_state = 12345
```
vishpillai123, line 12: Do lines 4-12 need to go under modeling.training? OR are they meant to be used for more than just modeling.training?

src/student_success_tool/configs/schemas/pdp_v2.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+    References:
+        - https://docs.databricks.com/en/machine-learning/automl/automl-api-reference.html#classify
+    """
+
+    exclude_cols: t.Optional[list[str]] = pyd.Field(
+        default=None,
+        description="One or more column names in dataset to exclude from training.",
+    )
+    time_col: t.Optional[str] = pyd.Field(
```
vishpillai123, line 123: This is interesting using a chronology column. How is this typically used?

src/student_success_tool/configs/schemas/pdp_v2.py
```diff
@@ -0,0 +1,249 @@
...
+    def check_some_nonnull_inputs(self):
+        if self.table_path is None and self.file_path is None:
+            raise ValueError("table_path and/or file_path must be non-null")
+        return self
+
+
+class DatasetConfig(pyd.BaseModel):
+    raw_course: DatasetIOConfig
+    raw_cohort: DatasetIOConfig
+    preprocessed: t.Optional[DatasetIOConfig] = None
```
vishpillai123, line 179: would preprocessed = training dataset?

vishpillai123 (approved):
This looks good! I had a few questions but overall I'm good with merging.

vishpillai123:
Oh also I wanted to ask though this is probably outside of the scope - if a school has multiple models (a retention and graduation model for example), would we want multiple configs or do we need to enable that within a config? 

I'm assuming multiple configs would be easier.

notebooks/pdp/config-v2-TEMPLATE.toml
bdewilde: Yup! These get used in both preprocessing and model training.

src/student_success_tool/configs/schemas/pdp_v2.py
```diff
@@ -0,0 +1,249 @@
...
+    )
+
+
+class ModelingConfig(pyd.BaseModel):
+    feature_selection: t.Optional[FeatureSelectionConfig] = None
+    training: TrainingConfig
+
+
+class InferenceConfig(pyd.BaseModel):
+    num_top_features: int = pyd.Field(default=5)
```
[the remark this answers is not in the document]
bdewilde: This is included with the trained model itself, see `min_prob_pos_label`

src/student_success_tool/configs/schemas/pdp_v2.py
bdewilde: In the case of the labeled dataset, yes. For unlabeled data, it would be the dataset to be given to the model as input in order to produce predictions. (Does that have a name?)

bdewilde:
> Oh also I wanted to ask though this is probably outside of the scope - if a school has multiple models (a retention and graduation model for example), would we want multiple configs or do we need to enable that within a config?
> 
> I'm assuming multiple configs would be easier.

Good question! But yes, out of scope -- it hasn't ever come for a PDP school, right? :)

I'd _probably_ follow the example of `datasets`, and make a top-level `models` dict with keys as model names and values as corresponding model configuration. I could do that as a fast follow-up, if you want.

PR #57, opened 2025-01-24
[pdp] Tweak project config + template + test case

bdewilde:
## changes

- changes `cfg.trained_model` into more flexible `cfg.models`, to allow for multiple named models trained using a given pipeline
- moves `min_prob_pos_label` from model to inference cfg, since that's _probably_ the only time in the process we'll need it; I'm not 100% sure on this tho, so I may move it back :)
- adds some descriptions to the config

## context

Inspired by feedback in PR #56


vishpillai123 (approved):
Looks good! Thanks for this update :)

PR #58, opened 2025-01-24
Modify load_feature_tables() function to allow for absolute & relative paths

vishpillai123:
I was running into issues with using this function and loading a TOML file from my local directory. 

## changes
- Added is_absolute() to resolve whether path is relative or absolute
- If referring to a path defined within module or an absolute path elsewhere, the function will retrieve the table. 

## context
- I ran into an issue using this function for one of my schools (not PDP).

## questions
None

### DIFF
--- a/src/student_success_tool/modeling/utils.py
+++ b/src/student_success_tool/modeling/utils.py
@@ -1,5 +1,4 @@
 import logging
-import os
 import pathlib
 import typing as t
 from collections.abc import Sequence
--- a/src/student_success_tool/modeling/utils.py
+++ b/src/student_success_tool/modeling/utils.py
@@ -86,27 +86,29 @@ def compute_sample_weights(
     )
 
 
-def load_features_table(rel_fpath: str) -> dict[str, dict[str, str]]:
+def load_features_table(fpath: str) -> dict[str, dict[str, str]]:
     """
     Load a features table mapping columns to readable names and (optionally) descriptions
-    from a TOML file located at ``rel_fpath`` within this package.
+    from a TOML file located at ``fpath``, which can either refer to a relative path in this
+    package or an absolute path loaded from local disk.
 
     Args:
-        rel_fpath: Path to features table TOML file relative to package root;
-            for example: "assets/pdp/features_table.toml".
-    """
-    try:
-        # If __file__ is defined, we're in a regular script or module
-        pkg_root_dir = next(
-            p
-            for p in pathlib.Path(__file__).parents
-            if p.parts[-1] == "student_success_tool"
-        )
-    except NameError:
-        # If __file__ is not available (i.e. Jupyter notebook), use the current working directory
-        pkg_root_dir = pathlib.Path(os.getcwd())
+        fpath: Path to features table TOML file relative to package root or absolute;
+            for example: "assets/pdp/features_table.toml" or "/path/to/features_table.toml".
 
-    file_path = pkg_root_dir / rel_fpath
+    Returns:
+    """
+    pkg_root_dir = next(
+        p
+        for p in pathlib.Path(__file__).parents
+        if p.parts[-1] == "student_success_tool"
+    )
+
+    file_path = pathlib.Path(fpath) if pathlib.Path(fpath).is_absolute() else pkg_root_dir / fpath
+
+    if not file_path.exists():
+        raise FileNotFoundError(f"The file at '{file_path}' could not be found.")
+
     with file_path.open(mode="rb") as f:
         features_table = tomllib.load(f)
     LOGGER.info("loaded features table from '%s'", file_path)
--- a/src/student_success_tool/modeling/utils.py
+++ b/src/student_success_tool/modeling/utils.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import pathlib
 import typing as t
 from collections.abc import Sequence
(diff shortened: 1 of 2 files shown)


src/student_success_tool/modeling/utils.py
```diff
@@ -94,11 +95,17 @@ def load_features_table(rel_fpath: str) -> dict[str, dict[str, str]]:
...
-    )
+    try:
+        # If __file__ is defined, we're in a regular script or module
+        pkg_root_dir = next(
+            p
+            for p in pathlib.Path(__file__).parents
+            if p.parts[-1] == "student_success_tool"
+        )
+    except NameError:
+        # If __file__ is not available (i.e. Jupyter notebook), use the current working directory
```
bdewilde, line 106: I'm very surprised by this! I thought `__file__` refers to the path to this module, no matter the env in which it's being imported. In fact I don't believe I've ever had this issue in a Notebook.

tests/modeling/test_utils.py
```diff
@@ -82,3 +87,58 @@ def test_compute_sample_weights(df, target_col, class_weight, exp):
...
+            FileNotFoundError,  
+        ),
+    ]
+)
+
+def test_load_features_table(tmpdir, toml_content, expected_output, expect_exception):
+    if toml_content:
+        toml_file = tmpdir.join("features_table.toml")
+        toml_file.write(toml_content)  
+
```
bdewilde, line 131: nit: remove whitespace line

tests/modeling/test_utils.py
```diff
@@ -82,3 +87,58 @@ def test_compute_sample_weights(df, target_col, class_weight, exp):
...
+        toml_file = tmpdir.join("features_table.toml")
+        toml_file.write(toml_content)  
+        
+        file_path = str(toml_file)
+    else:
+        file_path = "non_existent_path/features_table.toml"
+    
+    if expect_exception:
+        with pytest.raises(expect_exception):
+            utils.load_features_table(file_path)
```
bdewilde, line 138: In the case of "no error", standard practice is to use `from contextlib import nullcontext as does_not_raise` then `with pytest.raises(does_not_raise)` works in the same manner as an exception. You can find a couple examples of this pattern in other unit tests.

bdewilde (commented):
Can you show me an example of where the current code doesn't work?

vishpillai123:
> Can you show me an example of where the current code doesn't work?

Hi, so I'm trying to load a custom feature_table from my local directory (not for a PDP school), which is why I ran into this issue. But I also realized the `rel_fpath` is meant to refer to a path within the module.

Is there a way we can specify a path in our local directory? If not, how can we create that functionality? 

I'm not sure if this PR is really solving this specific issue, so we can either close this out or if you have an idea of how to address, you can share and I can implement.

image

bdewilde:
@vishpillai123 There's probably a "well-behaved" way to handle either case: The user can pass a _relative_ path to load files within the src code, or an _absolute_ path to load from a file wherever it may live on disk. Does that make sense? Can you implement that logic? Potentially useful: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_absolute

vishpillai123:
@bdewilde Made some updates based on your recommendation! Let me know what you think :). I also tested it for my school and it worked so this addresses my concerns. Previous functionality with relative path is the same.

src/student_success_tool/modeling/utils.py
```diff
@@ -85,21 +85,29 @@ class frequencies in the input data.
...
+    from a TOML file located at ``fpath``, which can either refer to a relative path in this
+    package or an absolute path loaded from local disk.
 
     Args:
-        rel_fpath: Path to features table TOML file relative to package root;
-            for example: "assets/pdp/features_table.toml".
+        fpath: Path to features table TOML file relative to package root or absolute;
+            for example: "assets/pdp/features_table.toml" or "/path/to/features_table.toml".
+
+    Returns:
```
bdewilde, line 98: nit: remove this?

src/student_success_tool/modeling/utils.py
```diff
@@ -85,21 +85,29 @@ class frequencies in the input data.
...
     """
     pkg_root_dir = next(
         p
         for p in pathlib.Path(__file__).parents
         if p.parts[-1] == "student_success_tool"
     )
-    file_path = pkg_root_dir / rel_fpath
+
```
bdewilde, line 105: nit: remove empty whitespace line

src/student_success_tool/modeling/utils.py
```diff
@@ -85,21 +85,29 @@ class frequencies in the input data.
...
+    file_path = pathlib.Path(fpath) if pathlib.Path(fpath).is_absolute() else pkg_root_dir / fpath
+
+    if not file_path.exists():
+        raise FileNotFoundError(f"The file at '{file_path}' could not be found.")
```
bdewilde, line 109: This is totally fine! But does this error not get raised automatically by the line below, `file_path.open(mode="rb")` ? If it does, we should drop the manual error-handling here, and leave it in Python's capable hands. :)

```python
In [1]: import pathlib

In [2]: file_path = pathlib.Path("foo/bar/bat.baz")

In [3]: with file_path.open(mode="rb") as f:
   ...:     _ = f.read()
   ...:
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
Cell In[3], line 1
----> 1 with file_path.open(mode="rb") as f:
      2     _ = f.read()

File ~/.pyenv/versions/3.11.7/lib/python3.11/pathlib.py:1044, in Path.open(self, mode, buffering, encoding, errors, newline)
   1042 if "b" not in mode:
   1043     encoding = io.text_encoding(encoding)
-> 1044 return io.open(self, mode, buffering, encoding, errors, newline)

FileNotFoundError: [Errno 2] No such file or directory: 'foo/bar/bat.baz'
```

tests/modeling/test_utils.py
```diff
@@ -82,3 +88,53 @@ def test_compute_sample_weights(df, target_col, class_weight, exp):
...
+        file_path = "non_existent_path/features_table.toml"
+    
+    with expect_exception:
+        features_table = utils.load_features_table(file_path)
+        
+        if expect_exception is does_not_raise():
+            assert isinstance(features_table, dict)
+            for key, value in expected_output.items():
+                assert key in features_table
+                assert features_table[key] == value
```
bdewilde, line 140: I'm not 100% sure, but I think pytest is smart enough to allow something like `assert features_table == expected_output`

bdewilde (approved):
Thanks for extending the functionality here! I don't think anything is "broken" here, so approving 👍  but there are some nips and tucks that could make the code+tests cleaner.

bdewilde:
I see you didn't make any of my suggested changes -- all good, but I think I'll do it for you ;)

PR #55, opened 2025-01-17
[pdp] Add inference nb template

bdewilde:

PR #59, opened 2025-01-25
[pdp] Refine data assessment template nb

bdewilde:
## changes

updates pdp data assessment template nb to leverage new project configs, fixes a few minor bugs, and adds more instruction in the middle and end wrt action items

## context

a subset of changes in PR #55, broken out for reviewability

### DIFF
--- a/notebooks/pdp/01-data-assessment-eda-TEMPLATE.py
+++ b/notebooks/pdp/01-data-assessment-eda-TEMPLATE.py
@@ -15,7 +15,7 @@
 # COMMAND ----------
 
 # MAGIC %md
-# MAGIC # Setup
+# MAGIC # setup
 
 # COMMAND ----------
 
@@ -45,14 +45,15 @@ from databricks.connect import DatabricksSession
 from databricks.sdk.runtime import dbutils
 
 from student_success_tool.analysis import pdp
+from student_success_tool import configs
 
 # COMMAND ----------
 
-logging.basicConfig(level=logging.INFO)
+logging.basicConfig(level=logging.INFO, force=True)
 logging.getLogger("py4j").setLevel(logging.WARNING)  # ignore databricks logger
 
 try:
-    spark_session = DatabricksSession.builder.getOrCreate()
+    spark = DatabricksSession.builder.getOrCreate()
 except Exception:
     logging.warning("unable to create spark session; are you in a Databricks runtime?")
     pass
@@ -60,7 +61,7 @@ except Exception:
 # COMMAND ----------
 
 # MAGIC %md
-# MAGIC ## `student-success-intervention` hacks
+# MAGIC ## import school-specific code
 
 # COMMAND ----------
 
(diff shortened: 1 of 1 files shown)


notebooks/pdp/01-data-assessment-eda-TEMPLATE.py
```diff
@@ -69,38 +69,23 @@
...
 from analysis import *  # noqa: F403
 
 # COMMAND ----------
 
-# MAGIC %md
-# MAGIC ## unity catalog config
-
-# COMMAND ----------
-
-catalog = "sst_dev"
```
vishpillai123, line 86: Why did we want to remove the `catalog`, `read_schema` and `write_schema` variables? Just curious because I am seeing the `CATALOG`, `INST_NAME_bronze`, and `INST_NAME_silver` strings below. 

I like having the variables and f-strings because I think it's less manual and not as much typing, but just wanted to understand the motivation of removing and having the users input directly into the different paths below.

notebooks/pdp/01-data-assessment-eda-TEMPLATE.py
```diff
@@ -509,15 +520,19 @@
...
 # MAGIC %md
 # MAGIC ### filter invalid rows(?)
 
 # COMMAND ----------
 
 # this is probably a filter you'll want to apply
 # these courses known to be an issue w/ PDP data
-df_course_valid = df_course.loc[df_course["course_number"].notna(), :]
-df_course_valid
+df_course_filtered = df_course.loc[df_course["course_number"].notna(), :]
```
vishpillai123, line 534: I like calling this filtered vs. valid!

notebooks/pdp/01-data-assessment-eda-TEMPLATE.py
```diff
@@ -574,6 +589,7 @@
 
 ax = sb.histplot(
     df_course.sort_values(by="academic_year"),
+    # df_course_filtered.sort_values(by="academic_year"),
```
vishpillai123, line 592: Thanks for adding this! I kept going back and forth with looking at `df_course` and `df_course_valid`, so I like adding this comment for each plot.

vishpillai123 (changes requested):
Hi this looks good for the most part, though i was curious why we wanted to remove the f-strings and variable names?

notebooks/pdp/01-data-assessment-eda-TEMPLATE.py
bdewilde: Looks like you realized why this was done later in the review -- project configs are replacing much of the hard-coded "magic" variables+logic. Will loop back round if I've misread your later comments.

notebooks/pdp/01-data-assessment-eda-TEMPLATE.py
bdewilde: no prob! always on the lookout for minor quality-of-life improvements

vishpillai123:
Got it, thanks for clarifying! This looks good to me!

vishpillai123 (approved):

PR #60, opened 2025-01-25
[pdp] Refine prep modeling dataset template nb

bdewilde:
## changes

updates pdp prep modeling dataset template nb to leverage new project configs, work for both "train" and "predict" runs, and adds more instruction wrt action items

## context

a subset of changes in PR #55, broken out for reviewability
sibling to PR #59 

## questions

- I've moved feature selection **out** of this step of the pipeline. For multiple reasons (I can elaborate), this works better and more safely alongside model training. Any strong objections to this?

### DIFF
--- a/notebooks/pdp/02-prepare-modeling-dataset-TEMPLATE.py
+++ b/notebooks/pdp/02-prepare-modeling-dataset-TEMPLATE.py
@@ -38,11 +38,11 @@ import functools as ft
 import logging
 import sys
 
-import matplotlib.pyplot
-import numpy as np
-import pandas as pd
+import matplotlib.pyplot as plt
+import pandas as pd  # noqa: F401
 import seaborn as sb
 from databricks.connect import DatabricksSession
+from databricks.sdk.runtime import dbutils
 from py4j.protocol import Py4JJavaError
 
 from student_success_tool import configs, modeling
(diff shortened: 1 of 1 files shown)


notebooks/pdp/02-prepare-modeling-dataset-TEMPLATE.py
```diff
@@ -192,125 +199,185 @@
...
-# drop unwanted columns and mask values by time
-df_labeled = pdp.dataops.clean_up_labeled_dataset_cols_and_vals(df_labeled)
-df_labeled.shape
+# MAGIC %md
+# MAGIC # STOP HERE?
+
+# COMMAND ----------
+
+# MAGIC %md
+# MAGIC Everything after this point is for labeled training data only. In case of a non-"train" run type, we'll just save the modeling dataset here and bail out of the notebook early.
```
vishpillai123, line 269: We can use dbutils.widgets() in DB workflow to specify the pipeline type here ("predict" vs. "train").

vishpillai123 (approved):
hey, I think this looks good and is definitely an upgrade with the use of "cfg". Makes the notebook more streamlined.

notebooks/pdp/02-prepare-modeling-dataset-TEMPLATE.py
bdewilde: I have it set up like this at the top of the nb:

```python
# check if we're running this notebook as a "job" in a "workflow"
# if not, assume this is a training workflow using labeled data
try:
    run_type = dbutils.widgets.get("run_type")
    dataset_name = dbutils.widgets.get("dataset_name")
except Py4JJavaError:
    run_type = "train"
    dataset_name = "labeled"
```

Is that what you're suggesting?

bdewilde: (going to merge, but happy to make follow-up changes if this isn't what you have in mind)

vishpillai123: Yes this is exactly what I meant! I think it would be good to have this on all of our templates, but I'm curious what you think as well.

PR #61, opened 2025-01-27
[pdp] Refine train+eval model template nb

bdewilde:
## changes

- updates pdp model train+evaluate template nb to leverage new project configs
- makes a couple col names in project config optional, with non-null defaults
- fixes a minor bug in automl wrapper function logic

## context

sibling to PRs #59 and #60 

## questions

- This nb is a little rougher than the preceding two because it hasn't _really_ been tested in actual pipelines yet. Do you see any serious problems or omissions? Happy to make this nb more robust here, at the same time as project config functionality is added in.

### DIFF
--- a/notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
+++ b/notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
@@ -23,8 +23,10 @@
 
 # COMMAND ----------
 
+# WARNING: AutoML/mlflow expect particular packages within certain version constraints
+# overriding existing installs can result in errors and inability to load trained models
 # install (minimal!) extra dependencies not provided by databricks runtime
-# %pip install "student-success-tool==0.1.0"
+# %pip install "student-success-tool==0.1.0" --no-deps
 # %pip install git+https://github.com/datakind/student-success-tool.git@develop --no-deps
 # %pip install pandera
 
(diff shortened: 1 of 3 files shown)


src/student_success_tool/modeling/training.py
```diff
@@ -68,7 +68,8 @@ def run_automl_classification(
     kwargs.setdefault("timeout_minutes", 5)
     exclude_cols = kwargs.pop("exclude_cols", [])
     assert isinstance(exclude_cols, list)  # type guard
-    if student_id_col is not None:
+    exclude_cols = exclude_cols.copy()
```
vishpillai123, line 71: This seems like a strange line to include. Why did we copy?

notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
```diff
@@ -127,57 +185,97 @@
...
+        )
+    )
     df_features = df.drop(columns=non_feature_cols)
-# TODO: figure this out
+# TODO(Burton): figure out if this is needed/wanted here
 # else:
-#     train_df = extract_training_data_from_model(experiment_id)
-#     train_df[prediction_col] = model.predict(train_df)
-#     train_df[risk_score_col] = model.predict_proba(train_df)[:, 1]
+#     df_train = modeling.evaluation.extract_training_data_from_model(experiment_id)
```
vishpillai123, line 265: This code looks very similar to what we have with the cell below (lines 275 - 278). Can we delete these comments?

vishpillai123 (approved):
The config cleans up a lot of the code and makes it look more structured. I just left two comments/thoughts, but overall this looks good so approving.

I am also doing work around bias assessment and using false negative parity rate & creating decisions based on statistical significance. Anyways, I may update this template once we add that functionality (in the coming weeks).

src/student_success_tool/modeling/training.py
bdewilde: I alluded to the issue in the commit message, but tl;dr: We were modifying this input list in-place, and that caused the input list in other contexts (_outside this function_) to be modified in ways that were bad. So, we create a copy, to avoid such badness.

notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
bdewilde: Let me double-check -- this code went through a lot of iterations, and I probably just wasn't 100% sure it was okay to part with this cruft.

PR #62, opened 2025-01-29
creating new branch for public repo contribution

nm3224:
## context & changes
- I moved this into the public repo since it makes more sense and allows me to leverage the existing functionalities in the `targets` file!
- `compute_target_variable` is modified in the `failure_to_graduate_within_x_years.py` file by labeling those with >4 year graduation or NA for that column as True (in need of support), else False (graduated within (<=4) years)
- There's a new functions in `shared.py` since the check-in for Rutgers is at the end of student's first year - I grabbed the second record, not the 1st. The intent is to only keep students who completed a full first-year (or 2 academic terms) since that's the check-in. It is called: `get_nth_student_terms`. It has an arg specifying whether that is within the cohort year, or outside of it. 

### IMPORTANT NOTE:
- **Should note somewhere to always confirm with the school what to do / how to treat `NaN` intensities in the `enrollment_intensity_first_term` column, b/c be default, this pipeline drops them! Should re-code them appropriately prior to running, _before_ using the `select_students_time_left` function, if seeking to include them in the analysis/modeling dataset!**


src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,142 @@
...
+def make_labeled_dataset(
+    df: pd.DataFrame,
+    *,
+    student_criteria: dict[str, object | Collection[object]],
+    student_id_cols: str | list[str] = "student_guid",
+    exclude_pre_cohort_terms: bool = True,
+    term_is_pre_cohort_col: str = "term_is_pre_cohort",
+    years_to_bach_at_cohort_col: str = "first_year_to_bachelors_at_cohort_inst",
+    max_years_to_graduation: int = 4,
+    intensity_time_lefts: list[tuple[str, float, t.Literal["year", "term"]]],
```
nm3224, line 20: re @bdewilde quoting you: 
"These arguments both represent the same thing, no? I'd recommend getting rid of max_years_to_graduation and using the more flexible data structure so you can specify max years differently between part- and full-time students." - yes this is a great idea and I will make that change soon!

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,142 @@
...
+        years_to_bach_at_cohort_col=years_to_bach_at_cohort_col,
+    )
+    df_labeled = pd.merge(df_features, df_targets, on=student_id_cols, how="inner")
+    return df_labeled
+
+
+def compute_target_variable(  
+    df: pd.DataFrame,  
+    *,  
+    student_id_cols: str | list[str] = "student_guid",
```
nm3224, line 76: re @bdewilde quoting you: "Does this not need to vary by enrollment intensity?"

As of now, we are treating them the same b/c we don't have an answer from Rutgers yet on what a part-time degree student timeline should be- so for now, outcome variable should be the same, evaluated with the cutoff of <=4 years, or >4 years. But in the future, that may be different - so you're right, I've changed that!

src/student_success_tool/analysis/pdp/targets/shared.py
```diff
@@ -280,6 +280,72 @@ def _compute_intensity_num_terms(
         for intensity, time, unit in intensity_time_lefts
     ]
 
+# THIS IS A NEW / MODIFIED FUNCTION TO GET THE TERM REPRESENTING END OF FIRST YEAR FOR A STUDENT, including pre-cohort cols
+
+def get_first_student_year_term(
```
nm3224, line 285: since the check-in for Rutgers is at the end of student's first year - I grabbed the second record, not the 1st. This is to ensure we're only keeping students who actually completed a full first-year (I'm assuming students who didn't wouldn't have a second record). And yes, we want the second term of the first year no matter what that term is. Does that make sense / seem sensible or am I missing something/another way to do it/misinterpreting the ask?

bdewilde: Hey, as I've said before, this function's name (and its docstring) is actively misleading: You're selecting the student's second term, regardless of whether the student had three terms in their first year, or if they had pre-cohort terms before their first year, or some other edge case. If you don't care about these cases, you should make that clear in the function name and docs: something like `get_nth_student_term(df, n, ...)`, and then you can use `n=1` for your use case. (Be sure to note that `n=0` => "first", etc.)

src/student_success_tool/analysis/pdp/targets/shared.py
```diff
@@ -280,6 +280,72 @@ def _compute_intensity_num_terms(
         for intensity, time, unit in intensity_time_lefts
     ]
 
+# THIS IS A NEW / MODIFIED FUNCTION TO GET THE TERM REPRESENTING END OF FIRST YEAR FOR A STUDENT, including pre-cohort cols
```
bdewilde, line 283: There's no need for code comments like this, please remove.

src/student_success_tool/analysis/pdp/targets/shared.py
```diff
@@ -280,6 +280,72 @@ def _compute_intensity_num_terms(
...
+    return (
+        df.loc[:, cols]
+        .sort_values(by=sort_cols, ascending=True)
+        .groupby(by=student_id_cols, sort=False, as_index=False)
+        .nth(1)
+    )
+
+# THIS IS A NEW / MODIFIED FUNCTION TO GET THE TERM REPRESENTING END OF FIRST YEAR FOR A STUDENT, excluding pre-cohort cols
+
+def get_first_student_year_term_within_cohort(
```
bdewilde, line 320: Seems like we should just bundle this functionality into the function this one is calling under the hood, maybe by adding a `exclude_pre_cohort_terms: bool` arg?

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,142 @@
...
+    )
+    df_labeled = pd.merge(df_features, df_targets, on=student_id_cols, how="inner")
+    return df_labeled
+
+
+def compute_target_variable(  
+    df: pd.DataFrame,  
+    *,  
+    student_id_cols: str | list[str] = "student_guid",  
+    years_to_bach_at_cohort_col: str = "first_year_to_bachelors_at_cohort_inst",
```
bdewilde, line 77: This arg's name assumes we'll always be using this for a bachelor's degree, which isn't necessarily the case. (Would be nice to use for Associate's!) Could you change the name to something like `years_to_degree_col`?

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,142 @@
...
+    """ 
+    student_id_cols = utils.to_list(student_id_cols)  
+    return (  
+    df[student_id_cols + [years_to_bach_at_cohort_col]]  
+    .drop_duplicates(subset=student_id_cols, ignore_index=True)  
+    .assign(target=lambda df: (df[years_to_bach_at_cohort_col] > max_years_to_graduation) | df[years_to_bach_at_cohort_col].isna())  
+    .fillna({"target": False})  
+    .astype({"target": "bool"})  
+    .set_index(student_id_cols)  
+    .loc[:, "target"]
```
bdewilde, line 94: style: fix your whitespace please!

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,142 @@
...
+    Args:
+        df
+        student_id_cols
+        years_to_bach_at_cohort_col
+    """ 
+    student_id_cols = utils.to_list(student_id_cols)  
+    return (  
+    df[student_id_cols + [years_to_bach_at_cohort_col]]  
+    .drop_duplicates(subset=student_id_cols, ignore_index=True)  
+    .assign(target=lambda df: (df[years_to_bach_at_cohort_col] > max_years_to_graduation) | df[years_to_bach_at_cohort_col].isna())
```
bdewilde, line 90: `max_years_to_graduation` isn't a function argument -- where is this coming from?

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,142 @@
...
+    df_labeled = pd.merge(df_features, df_targets, on=student_id_cols, how="inner")
+    return df_labeled
+
+
+def compute_target_variable(  
+    df: pd.DataFrame,  
+    *,  
+    student_id_cols: str | list[str] = "student_guid",  
+    years_to_bach_at_cohort_col: str = "first_year_to_bachelors_at_cohort_inst",  
+    intensity_time_lefts: list[tuple[str, float, t.Literal["year", "term"]]],
```
bdewilde, line 78: It looks like you never use this argument in the function's logic. Why's it here then?

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,142 @@
...
+        student_id_cols
+        years_to_bach_at_cohort_col
+    """ 
+    student_id_cols = utils.to_list(student_id_cols)  
+    return (  
+    df[student_id_cols + [years_to_bach_at_cohort_col]]  
+    .drop_duplicates(subset=student_id_cols, ignore_index=True)  
+    .assign(target=lambda df: (df[years_to_bach_at_cohort_col] > max_years_to_graduation) | df[years_to_bach_at_cohort_col].isna())  
+    .fillna({"target": False})  
+    .astype({"target": "bool"})
```
bdewilde, line 92: nit: Use `pandas` "boolean" dtype, which is more flexible than stdlib `bool`

```suggestion
    .astype({"target": "boolean"})  
```

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,142 @@
...
+        df
+        student_id_cols
+        years_to_bach_at_cohort_col
+    """ 
+    student_id_cols = utils.to_list(student_id_cols)  
+    return (  
+    df[student_id_cols + [years_to_bach_at_cohort_col]]  
+    .drop_duplicates(subset=student_id_cols, ignore_index=True)  
+    .assign(target=lambda df: (df[years_to_bach_at_cohort_col] > max_years_to_graduation) | df[years_to_bach_at_cohort_col].isna())  
+    .fillna({"target": False})
```
bdewilde, line 91: In what cases does the above produce null values that we need to fill with False?

bdewilde (changes requested):

bdewilde:
Does this do what we need, in the general case that different enrollment intensities have different time limits for graduating "on time"?

```python
def compute_target_variable(
    df: pd.DataFrame,
    *,
    intensity_time_limits: dict[str, int],
    student_id_cols: str | list[str] = "student_guid",
    enrollment_intensity_col: str = "enrollment_intensity_first_term",
    years_to_degree_col: str = "first_year_to_bachelors_at_cohort_inst",
) -> pd.Series:
    """
    Args:
        df
        intensity_time_limits: Mapping of enrollment intensity value (e.g. "FULL-TIME")
            to the maximum number of years allowed to earn degree "in time" (e.g. 4).
        student_id_cols
        enrollment_intensity_col
        years_to_degree_col: Name of column giving the year in which students _first_
            earned a particular degree (by default, a Bachelor's degree).
    """
    student_id_cols = utils.to_list(student_id_cols)
    include_cols = student_id_cols + [enrollment_intensity_col, years_to_degree_col]
    # we want a target for every student in input df; this will ensure it
    df_distinct_students = df[student_id_cols].drop_duplicates(ignore_index=True)
    df_first_terms = shared.get_first_student_terms(
        df, student_id_cols=student_id_cols, include_cols=include_cols
    )
    # compute all intensity/year boolean arrays separately
    # then combine with a logical OR
    targets = [
        (
            # enrollment intensity is equal to a specified value
            df_first_terms[enrollment_intensity_col].eq(intensity)
            & (
                # they graduated after max num years allowed
                (df_first_terms[years_to_degree_col]).gt(num_years)
                # or they *never* graduated
                | df_first_terms[years_to_degree_col].isna()
            )
        )
        for intensity, num_years in intensity_time_limits.items()
    ]
    target = np.logical_or.reduce(targets)
    df_target_true = (
        df_first_terms.loc[target, student_id_cols]
        .assign(target=True)
    )
    return (
        # all students not assigned True, now assigned False
        pd.merge(df_distinct_students, df_target_true, on=student_id_cols, how="left")
        .fillna(False)
        .astype({"target": "bool"})
        .set_index(student_id_cols)
        .loc[:, "target"]
    )
```

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
bdewilde: Update: Nevermind, I think I remember actually _disfavoring_ the "boolean" dtype for compatibility reasons in AutoML, though it only works if there are no null values in the "target" column.

src/student_success_tool/analysis/pdp/targets/shared.py
nm3224: done!

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
nm3224: none - we are assuming nulls = non grads, these students are all filtered to have had atleast 4 years to graduate

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,192 @@
...
+            .drop_duplicates(subset=student_id_cols, ignore_index=True)    
+            .assign(target=lambda df: (df[years_to_degree_col].gt(df[enrollment_intensity_col].map(intensity_num_terms))) |   
+                                     (df[years_to_degree_col].isna()))    
+            .fillna({"target": False})    
+            .astype({"target": "bool"})    
+            .set_index(student_id_cols)    
+            .loc[:, "target"]    
+            )    
+
+def select_eligible_students(
```
nm3224, line 115: @bdewilde mind taking a look at the logic here and letting me know what you think?

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,192 @@
...
+        student_id_cols  
+        years_to_degree_col  
+        intensity_time_lefts  
+        intensity_col  
+    """   
+    student_id_cols = utils.to_list(student_id_cols)  
+    intensity_num_terms = {intensity: time if unit == "year" else time / num_terms_in_year  
+                           for intensity, time, unit in intensity_time_lefts  
+                           }   
+    return (
```
nm3224, line 104: @bdewilde and here too please! I tried what you had wrote, it was a little buggy and wasn't labeling correctly. I think that may have been for other reasons that I've now fixed so I'll test it again in a bit - but this current code is tested and works!

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
nm3224: removed this - it's not necessary

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
bdewilde: I'm reasonably confident in what I wrote, since it was adapted from existing code that definitely does work -- can you be more specific about what was "buggy"? It's hard to respond effectively to prompts like "let me know what you think", I need a little more specificity in what you're wondering about.

src/student_success_tool/analysis/pdp/targets/shared.py
```diff
@@ -280,6 +280,74 @@ def _compute_intensity_num_terms(
...
+        else list(
+            utils.unique_elements_in_order(student_id_cols + sort_cols + include_cols)
+        )
+    )
+    return (
+        df.loc[:, cols]
+        .sort_values(by=sort_cols, ascending=True)
+        .groupby(by=student_id_cols, sort=False, as_index=False)
+        .nth(n)
+    )
```
bdewilde, line 316: looks correct to me 👍

src/student_success_tool/analysis/pdp/targets/shared.py
```diff
@@ -280,6 +280,74 @@ def _compute_intensity_num_terms(
...
+) -> pd.DataFrame:
+    """
+    For each student, get the nth row in ``df`` (in ascending order of ``sort_cols`` )
+    and a configurable subset of columns. 
+
+    Args:
+        df
+        student_id_cols
+        sort_cols
+        n
```
bdewilde, line 299: You should document what `n` is here -- notably, that `n=0` => get the first term, i.e. equivalent to the functionality of `get_first_student_terms()`.

src/student_success_tool/analysis/pdp/targets/shared.py
```diff
@@ -280,6 +280,74 @@ def _compute_intensity_num_terms(
...
+    )
+    return (
+        df.loc[:, cols]
+        .sort_values(by=sort_cols, ascending=True)
+        .groupby(by=student_id_cols, sort=False, as_index=False)
+        .nth(n)
+    )
+
+
+def get_nth_student_terms_within_cohort(
```
bdewilde, line 319: I understand that you're repeating a pattern I started elsewhere in the code; however, I think that pattern is meh, and it only happened through historical accident. (The "within-cohort" variation was implemented long after the original.) We can leave this for now, but do ourselves a favor and add a TODO: combine with `get_nth_student_terms()` function by adding within-cohort args to configure behavior.

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,208 @@
...
+        shared.select_students_by_time_left(
+            df_ref,
+            student_id_cols=student_id_cols,
+            intensity_time_lefts=intensity_time_lefts,
+            max_term_rank=max_term_rank,
+            num_terms_in_year=num_terms_in_year,
+            enrollment_intensity_col=enrollment_intensity_col,
+            term_rank_col=term_rank_col
+        )
+    )
```
bdewilde, line 200: Looks like you have this same code twice in a row? Please remove one of them.

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,208 @@
...
+    df: pd.DataFrame,
+    *,
+    student_criteria: dict[str, object | Collection[object]],
+    student_id_cols: str | list[str] = "student_guid",
+    intensity_time_lefts: list[tuple[str, float, t.Literal["year", "term"]]],
+    max_term_rank: int,
+    num_terms_in_year: int = 4,
+    exclude_pre_cohort_terms: bool = True,
+    term_is_pre_cohort_col: str = "term_is_pre_cohort",
+    n: int,
```
bdewilde, line 124: nit: move this arg up above the `*_col` args w/ defaults

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,208 @@
...
+            num_terms_in_year=num_terms_in_year,
+            enrollment_intensity_col=enrollment_intensity_col,
+            term_rank_col=term_rank_col
+        )
+    )
+    df_out = features.shared.merge_many_dataframes(
+        [df_students_by_criteria, df_students_by_checkin, df_students_by_time_left],
+        on=student_id_cols,
+        how="inner",
+    )
```
bdewilde, line 205: Making sure I understand: You're selecting eligible students by "criteria", by having at least "time left" from first term, and by being enrolled in an nth term? If that's right, could you describe that in the function's docstring?

I recognize that you've used the `failure_to_earn_enough_credits_in_time_from_enrollment` code as your reference example, and that does not have a suitable docstring. Please don't follow that bad example! I was super time-pressed when I wrote that code, and haven't yet gotten round to making it better.

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,208 @@
...
+        df  
+        student_id_cols  
+        years_to_degree_col  
+        intensity_time_lefts  
+        intensity_col  
+    """   
+    student_id_cols = utils.to_list(student_id_cols)  
+    intensity_num_terms = {intensity: time if unit == "year" else time / num_terms_in_year  
+                           for intensity, time, unit in intensity_time_lefts  
+                           }
```
bdewilde, line 103: I think there's already a function for this: `shared._compute_intensity_num_terms(intensity_time_limits, num_terms_in_year)` , no? There's also `dataops.infer_num_terms_in_year(df[term_col])`

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,208 @@
...
+                           }   
+    return (    
+            df[student_id_cols + [years_to_degree_col, enrollment_intensity_col]]    
+            .drop_duplicates(subset=student_id_cols, ignore_index=True)    
+            .assign(target=lambda df: (df[years_to_degree_col].gt(df[enrollment_intensity_col].map(intensity_num_terms))) |   
+                                     (df[years_to_degree_col].isna()))       
+            .astype({"target": "bool"})    
+            .set_index(student_id_cols)    
+            .loc[:, "target"]    
+            )
```
bdewilde, line 112: Could you format this code using `ruff format` ? It's hard to read with the whitespace as it is 🙏

src/student_success_tool/analysis/pdp/targets/shared.py
nm3224: @bdewilde do you want to scrap the get_first_student_terms function and just utilize this with a default n=0 instead since it's universal / more flexible?

src/student_success_tool/analysis/pdp/targets/shared.py
nm3224: yes agreed!

src/student_success_tool/analysis/pdp/targets/shared.py
bdewilde: We could consolidate functionality under the hood at some point, but not here in this PR, and it's probably not necessary anyway -- pandas has both `.first()` and `.nth()` for a reason :)

src/student_success_tool/analysis/pdp/targets/shared.py
```diff
@@ -312,7 +312,7 @@ def get_nth_student_terms(
             utils.unique_elements_in_order(student_id_cols + sort_cols + include_cols)
         )
     )
-    if exclude_pre_cohort_terms.eq(True):
+    if exclude_pre_cohort_terms == True:
```
bdewilde, line 315: good catch

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
nm3224: for some reason when i run it using the inference script it's funky. So i stuck to manual input

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
nm3224: How do i run this in terminal in databricks?

nm3224: nevermind figured it out!

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
bdewilde: Raising my eyebrows at this -- sounds like user error / maybe a bug in the calling code. Probably worth investigating!

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
bdewilde: You were able to run this in the terminal in databricks? Or you realized that this is the public repo, so you can run it locally via your text editor?

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
nm3224: It needed a few tweaks but it's good now!

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
nm3224: I'll look into it! The reason I have code for the `intensity_num_terms` is because what's in the shared transforms it from _years to terms_, but not _terms to years_. I wanted to account for the scenario that the user input is in terms, not years.

src/student_success_tool/analysis/pdp/targets/shared.py
```diff
@@ -281,6 +283,55 @@ def _compute_intensity_num_terms(
     ]
 
 
+def get_nth_student_terms(
+    df: pd.DataFrame,
+    *,
+    student_id_cols: str | list[str] = "student_guid",
+    n: int = 1,
```
bdewilde, line 290: I recommend moving `n` up above the cols args, and _not_ giving it a default value -- i.e. making it a required arg. I feel pretty strongly that choosing `n` should always be explicit.

src/student_success_tool/analysis/pdp/targets/shared.py
```diff
@@ -281,6 +283,55 @@ def _compute_intensity_num_terms(
...
+    student_id_cols = utils.to_list(student_id_cols)
+    sort_cols = utils.to_list(sort_cols)
+    cols = (
+        df.columns.tolist()
+        if include_cols is None
+        else list(
+            utils.unique_elements_in_order(student_id_cols + sort_cols + include_cols)
+        )
+    )
+    if exclude_pre_cohort_terms == True:
```
bdewilde, line 318: `is True` and `is False` (and `is None`) are idiomatic Python conditions -- don't ever use `== True`, unless a package specifically wants that construction in certain contexts (e.g. `sqlalchemy`, `numpy`).

```suggestion
    if exclude_pre_cohort_terms is True:
```

src/student_success_tool/analysis/pdp/targets/shared.py
```diff
@@ -281,6 +283,55 @@ def _compute_intensity_num_terms(
...
+        return (
+            # exclude rows that are "pre-cohort", so "first" meets our criteria here
```
bdewilde, line 320:
```suggestion
            # exclude rows that are "pre-cohort", so "nth" meets our criteria here
```

src/student_success_tool/analysis/pdp/targets/shared.py
```diff
@@ -281,6 +283,55 @@ def _compute_intensity_num_terms(
...
+            .groupby(by=student_id_cols, sort=False, as_index=False)
+            .nth(n)
+        )
+    else:
+        return (
+            df.loc[:, cols]
+            .sort_values(by=sort_cols, ascending=True)
+            .groupby(by=student_id_cols, sort=False, as_index=False)
+            .nth(n)
+        )
```
bdewilde, line 332: nit: You could make this just _a little_ DRYer, like so:

```python
df = (
    df.loc[df[term_is_pre_cohort_col].eq(False), cols]
    if exclude_pre_cohort_terms is True
    else df.loc[:, cols]
)
return (
    df.sort_values(by=sort_cols, ascending=True)
    .groupby(by=student_id_cols, sort=False, as_index=False)
    .nth(n)
)
```

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
```diff
@@ -0,0 +1,213 @@
...
+        df
+        intensity_time_limits: Mapping of enrollment intensity value (e.g. "FULL-TIME")
+            to the maximum number of years allowed to earn degree "in time" (e.g. 4).
+        student_id_cols
+        enrollment_intensity_col
+        years_to_degree_col: Name of column giving the year in which students _first_
+            earned a particular degree (by default, a Bachelor's degree).
+        num_terms_in_year: Defined number of terms in one academic year.
+    """
+    student_id_cols = utils.to_list(student_id_cols)
```
bdewilde, line 94: It looks like you landed on something close to what I suggested -- just checking, was that intentional? Last I heard, you found it buggy! :P

bdewilde (approved):
Sorry this took so long to review 😮‍💨 There are a few small things to update yet, but I think it's safe to merge after that.

As I've mentioned, I'm going to refactor this part of the code, such that targets are decoupled from "checkpoints", allowing for more flexibility in usage. Not something you need to worry about here, but just a heads-up.

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
nm3224: I had to fix a few things in it, and then I think once I fixed some of the other functions I got it to work!

src/student_success_tool/analysis/pdp/targets/failure_to_graduate_within_x_years.py
nm3224: in the terminal in databricks :)

PR #64, opened 2025-01-31
[pdp] Add inference template nb

bdewilde:
## changes

- adds pdp model predict + explain template nb consistent with preceding steps of the pipeline
- adds a utility function to load mlflow models from _the registry_
- adds a hacky plot to model eval nb

## context

final (and original intended) component of PR #55

## questions

please see the various TODOs embedded within the nb

### DIFF
--- a/notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
+++ b/notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
@@ -42,6 +42,7 @@ import os
 import mlflow
 import numpy as np
 import pandas as pd
+import sklearn.inspection
 import sklearn.metrics
 from databricks.connect import DatabricksSession
 from databricks.sdk.runtime import dbutils
@@ -378,7 +379,26 @@ with mlflow.start_run(run_id=run_id) as run:
 
 # COMMAND ----------
 
-# MAGIC %md
-# MAGIC TODO: See about adding permutation importance and/or global SHAP feature importance evaluation here
+# TODO TODO TODO
+result = sklearn.inspection.permutation_importance(
+    model,
+    df_features.drop(columns=cfg.target_col),
+    df_features[cfg.target_col],
+    scoring=sklearn.metrics.make_scorer(
+        sklearn.metrics.log_loss, greater_is_better=False
+    ),
+    n_repeats=10,
+)
+
+sorted_importances_idx = result.importances_mean.argsort()
+importances = pd.DataFrame(
+    result.importances[sorted_importances_idx].T,
+    columns=df_features.columns[sorted_importances_idx],
+)
+ax = importances.plot.box(vert=False, whis=10, figsize=(10, 10))
+ax.set_title("Permutation Importances (test set)")
+ax.axvline(x=0, color="k", linestyle="--")
+ax.set_xlabel("Decrease in accuracy score")
+ax.figure.tight_layout()
 
 # COMMAND ----------
--- a/notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
+++ b/notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
@@ -42,8 +42,6 @@ import mlflow
 import numpy as np
 import pandas as pd
 import shap
-import sklearn.inspection
-import sklearn.metrics
 from databricks.connect import DatabricksSession
 from databricks.sdk.runtime import dbutils
 from py4j.protocol import Py4JJavaError
@@ -129,13 +127,8 @@ df.head()
 
 # COMMAND ----------
 
+
 # TODO: get this functionality into public repo's modeling.inference?
-import typing as t
-
-import numpy as np
-import pandas as pd
-
-
 def predict_proba(
     X,
     model,
@@ -143,8 +136,7 @@ def predict_proba(
     feature_names: t.Optional[list[str]] = None,
     pos_label: t.Optional[bool | str] = None,
 ) -> np.ndarray:
-    """
-    """
+    """ """
     if feature_names is None:
         feature_names = model.named_steps["column_selector"].get_params()["cols"]
     if not isinstance(X, pd.DataFrame):
@@ -156,6 +148,7 @@ def predict_proba(
         return pred_probs[:, model.classes_.tolist().index(pos_label)]
     else:
         return pred_probs
+
 
 # COMMAND ----------
 
@@ -254,7 +247,10 @@ df_ref.shape
 
 explainer = shap.explainers.KernelExplainer(
     ft.partial(
-        predict_proba, model=model, feature_names=model_feature_names, pos_label=cfg.pos_label
+        predict_proba,
+        model=model,
+        feature_names=model_feature_names,
+        pos_label=cfg.pos_label,
     ),
     df_ref,
     link="identity",
@@ -269,7 +265,9 @@ shap_schema = StructType(
 )
 
 df_shap_values = (
-    spark.createDataFrame(df_test.reindex(columns=model_feature_names + [cfg.student_id_col]))  # noqa: F821
+    spark.createDataFrame(
+        df_test.reindex(columns=model_feature_names + [cfg.student_id_col])
+    )  # noqa: F821
     .repartition(sc.defaultParallelism)  # noqa: F821
     .mapInPandas(
         ft.partial(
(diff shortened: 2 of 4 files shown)


notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
```diff
@@ -0,0 +1,332 @@
...
+# to MLFlow experiments... which we don't want
+mlflow.autolog(disable=True)
+
+# COMMAND ----------
+
+# check if we're running this notebook as a "job" in a "workflow"
+# if not, assume this is a prediction workflow
+try:
+    run_type = dbutils.widgets.get("run_type")
+    dataset_name = dbutils.widgets.get("dataset_name")
```
vishpillai123, line 77: For the `dataset_name`, this is something you'll have to assign in the notebook before this: '[02-prepare-modeling-dataset-TEMPLATE.py](https://github.com/datakind/student-success-tool/blob/develop/notebooks/pdp/02-prepare-modeling-dataset-TEMPLATE.py)'. So for example, we can do this at the end of the notebook:

` dbutils.jobs.taskValues.set(key="dataset_name", value=dataset_name) # noqa: F821`

where dataset_name points to a dataframe etc.

We don't need to do this for `run_type` since this will be pulled down as one of the job parameters.

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
```diff
@@ -0,0 +1,332 @@
...
+    model_name = dbutils.widgets.get("model_name")
```
vishpillai123, line 78: What does the `model_name` represent here and how are we planning on retrieving this info? Would you want to be assigned via AutoML artifacts?

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
```diff
@@ -0,0 +1,332 @@
...
+        spark_session=spark,
+    )
+)
+df.head()
+
+# COMMAND ----------
+
+
+# TODO: get this functionality into public repo's modeling.inference?
+def predict_proba(
```
vishpillai123, line 132: Perhaps, this can go in `modeling.utils`... what do you think? Since this notebook is on the longer side.

src/student_success_tool/modeling/utils.py
```diff
@@ -110,3 +111,32 @@ def load_features_table(fpath: str) -> dict[str, dict[str, str]]:
...
+        - https://mlflow.org/docs/latest/python_api/mlflow.pyfunc.html#mlflow.pyfunc.load_model
+    """
+    load_model_func = (
+        mlflow.sklearn.load_model
+        if framework == "sklearn"
+        else mlflow.xgboost.load_model
+        if framework == "xgboost"
+        else mlflow.lightgbm.load_model
+        if framework == "lightgbm"
+        else mlflow.pyfunc.load_model
```
vishpillai123, line 138: I like this logic with loading based on model type, though in the final else case... I've personally had some issues with `mlflow.pyfunc.load_model`. Also when would we fall into this issue? I thought decision tree, logreg would be sklearn?

vishpillai123 (changes requested):
Hi @bdewilde, this is in great shape overall, but I just left some suggestions/comments.

notebooks/pdp/03-train-evaluate-model-TEMPLATE.py
```diff
@@ -378,7 +379,26 @@
 
 # COMMAND ----------
 
-# MAGIC %md
-# MAGIC TODO: See about adding permutation importance and/or global SHAP feature importance evaluation here
+# TODO TODO TODO
+result = sklearn.inspection.permutation_importance(
```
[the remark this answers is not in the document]
bdewilde: This is an excellent way to get _global_ feature importances, in a way that's performant (i.e. negligible in this context) and more statistically robust than getting importances from the model's coefficients or importances attributes. This isn't a replacement for SHAP, it's a complement.

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
bdewilde: How does one get `run_type` in the code, if not via this call?

I'm still a bit unsure about the best way to get parameters from a databricks "job" vs manually specifying. Here, I'm just following the example set in prior nbs.

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
bdewilde: `model_name` is analogous to `dataset_name` in that both are keys in the project config's `models` and `datasets` parameters, respectively.

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
bdewilde: I went back and forth on it... For now, I'd prefer to leave it here, with the TODO reminding us to clean it up later :)

src/student_success_tool/modeling/utils.py
bdewilde: Just covering the usual bases here. Strangely, it seemed like lightgbm and xgboost models were actually being saved using the "sklearn" framework, possibly through sklearn's built-in wrappers for those external libs, but I didn't delve too deeply. This function should work regardless if you give it the proper framework!

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
vishpillai123: run_type is defined in the job parameters. dataset_name can be defined in the prior notebook when you create the dataset. Then you can define this as one of the task parameters. You can check out some of the workflows that I shared to get some examples of both of these.

Screenshot 2025-02-04 at 12 09 14 PM

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
vishpillai123: Oh I see what you mean. Okay yeah we can refer to the config instead of `dbutils.widgets` here for at least the `model_name`. I think with `dataset_name`, it makes more sense to me to have that as a task parameter given that this will depend on whether we are training vs. predicting on a dataset (the path will be different).

vishpillai123: I would say for the sake of this PR, we can have `model_name` just pulled from the config. Are you okay with changing that?

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
bdewilde: Setting aside which parameters need to be set and in what form, the big question for me is how are we supposed to get and then leverage these parameters in notebooks. Could you catch me up?

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
bdewilde: This will require a change to the project config, so I'd prefer to punt that to a separate PR / chunk of work. Okay with you?

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
bdewilde: This source [here](https://stackoverflow.com/questions/69009013/how-do-we-access-databricks-job-parameters-inside-the-attached-notebook) suggests that `dbutils.widgets.get()` _is_ the correct way to get job parameters in a notebook. Is that not so?

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
vishpillai123: Yes, that's fair! I'll approve then and we can adjust later.

vishpillai123 (approved):

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
bdewilde: I have an idea for how this will work 👍

notebooks/pdp/04-make-explain-predictions-TEMPLATE.py
bdewilde: various confusions resolved _elsewhere_

We'll work out the kinks here in subsequent work.

PR #66, opened 2025-02-06
[pdp] add funcs to read raw datasets from table or file

bdewilde:
## changes

adds dataio functions to read raw pdp course+cohort datasets from file OR table, as a more general replacement for existing functions that only read these datasets from file

**note:** i'll remove the superseded functions in a later package release, to avoid breaking changes

## context

based on plans for running prediction pipelines, it looks like we want to start reading raw datasets from tables rather than files. that's fine! the data _should_ be the same (modulo any weirdness in how things get formatted -- daniel is working on standardizing this afaik). these functions enable that process, while also supporting existing processes.

### DIFF
--- a/src/student_success_tool/analysis/pdp/dataio.py
+++ b/src/student_success_tool/analysis/pdp/dataio.py
@@ -13,6 +13,190 @@ from . import utils
 LOGGER = logging.getLogger(__name__)
 
 
+def read_raw_pdp_course_data(
+    *,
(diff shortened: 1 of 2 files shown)


bdewilde:
Hey you two, since this is strictly additive (no existing functionality affected), I'm going to merge it in so I can move on to cutting a v0.1.1 release of this package. Sorry about the review fake-out!

PR #67, opened 2025-02-06
fix: ensure student guids are unique in raw data generation

ZakMiller:
Previously, there was no guarantee that the ids were unique. Given large enough numbers there would be collisions (which isn't allowed by the Schema).

This generates the ids up front using .unique so that there won't be collisions.

Moving the comment is something that ruff did automatically (the last line changed).


src/student_success_tool/generation/pdp/raw_cohort.py
```diff
@@ -25,7 +26,9 @@ def raw_cohort_record(
 
         # TODO: handle other cases, e.g. gateway course attempted/completed/grades
         record = {
-            "Student GUID": self.student_guid(),
+            "Student GUID": student_guid
```
ZakMiller, line 29: An alternative possible solution would be to use ".unique" here -> `self.unique.student_guid()`

That didn't work, though. My python isn't super strong and so I'm not exactly sure why. I guess when you add a provider that provider doesn't inherit all of the faker's properties?

bdewilde: Ha, I was going to suggest exactly this! I'm pretty sure there _is_ a way; maybe it's `self.generator.unique.student_guid()`? I've not had my head in `faker` in a while, and its eccentricities never stick...

ZakMiller: Good idea, just tried it, but that doesn't work either.
 Another option would be to write my own unique checking in the provider (wouldn't be that hard). We could also bump up the number of digits by a bunch but that wouldn't permenantly solve the problem or anything

bdewilde:
Hey @ZakMiller , just following up on this. What's your preference: Should we go with your first attempt, or try to find a "more correct" (heavy emphasis on the air quotes) way of doing this in faker?

ZakMiller:
> Hey @ZakMiller , just following up on this. What's your preference: Should we go with your first attempt, or try to find a "more correct" (heavy emphasis on the air quotes) way of doing this in faker?

I could be wrong but I'm not seeing a way of leveraging the built in unique stuff (after some experimentation) from within the provider so I think the options are my initial choice or rolling my own unique logic in the provider which would simplify the interface but make the implementation more complex. Personally I am leaning a little toward my first attempt but I'm happy to do either.

bdewilde (approved):
Let's go with what you've got now, and maybe revisit in the future if it ever bugs us.

Only thing to do is run ruff autoformatting over one of the files, so the "style" check passes: `uv tool run ruff format --check scripts/generate_synthetic_pdp_datasets.py src/student_success_tool/generation/pdp/raw_cohort.py`

PR #68, opened 2025-02-06
Adjust year logic to account for leap days

ZakMiller:
The previous code would occasionally return an error "day is out of range for month" when the randomly selected date was February 29th in a year that has a leap day (like 2024) (as February 29th of the following year would be invalid).

### DIFF
--- a/src/student_success_tool/generation/pdp/raw_cohort.py
+++ b/src/student_success_tool/generation/pdp/raw_cohort.py
@@ -157,13 +157,17 @@ class Provider(BaseProvider):
     def institution_id(self) -> str:
         return self.numerify("#####!")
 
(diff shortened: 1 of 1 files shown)


bdewilde (approved):
Coding with time is just _the worst_ 😪 

This is fine except in the edgiest of edge cases. If we _really_ wanted to be safe, we could do this instead:

```python
start_yr = start_dt.year
end_yr = start_yr + 1
return f"{start_yr}-{end_yr}"
```

ZakMiller:
Oh I guess this would be the same year if it was January 1st on a leap year?

ZakMiller:
I went with your solution and added a comment as well.

bdewilde:
@ZakMiller I think you just need to remove the now-unneeded timedelta import

ZakMiller:
Also updated pr name and commit msg

PR #69, opened 2025-02-07
Prepare v0.1.1 release changelog

bdewilde:
## changes

- sets package version to v0.1.1: it's _mostly_ compatible with v0.1.0, and this is still early days, so no need to bump to v0.2.0 until notable breaking changes occur
- adds changelog for this new version

## context

we're due for a version bump

## questions

- are there any reasons _not_ to cut this release?


vishpillai123 (approved):
Looks great! clear summary on recent work

PR #73, opened 2025-02-13
[pdp] refactor templates, add preprocess

bdewilde:
## changes

- moves and renames the "notebooks/pdp" templates so that it more closely resembles an actual school-specific pipeline directory
- adds example (no-op) school-specific overrides for base raw data schemas
-  adds "hack" `preprocess_data()` function that takes converted+validated course/cohort data and outputs a modeling dataset, encapsulating most of the functionality included in the 01-preprocess-data notebook
- updates imports and calls to reflect the recently refactored src code

## context

Now that we've settled on the structure for a school-specific pipeline, we should reflect that here as an example to follow. Additionally (iiuc) the Google folks wanted to see an example of what a school-specific preprocess function could look like.

## questions

- Google folks: Is that `preprocess_data()` function what you wanted to see? If not, could you give me some direction on what you'd like? Happy to do it, just unsure what you're looking for.
- Everyone: Any concerns about the restructuring / renaming?


vishpillai123 (approved):
I like the renaming. nice job adding one preprocessing func for the fellows.

PR #76, opened 2025-02-15
[pdp] add refactored targets

bdewilde:
## changes

- adds new-and-improved implementations for the three types of targets we currently support for pdp schools, with better functionality, consistent interfaces, and decoupled scope from filtering/checkpointing
- adds thorough unit tests for these new targets (this is a big deal!!)

**out-of-scope:**
- removing the old target modules
- adding subpackages for filtering and checkpointing student-term data

## context

Currently, "targets" are tightly coupled to student filtering and/or checkpoint selection, which is awkward at best and makes developing new combinations of target/filter/checkpoint logic much more difficult than it should be.

## questions

- how do you feel about separating target computation from student(-term) filtering and/or checkpointing?
- for the graduation target, i've added functionality such that the mode (most common) enrollment intensity across all terms prior to graduation is used to determine if the student graduated on time, rather than the first-term enrollment intensity. this is because students regularly change their intensity over the course of their academic journey, and we want to more accurately account for that reality. it's possible to recover the old behavior by choosing `"enrollment_intensity_first_term"` as the target's `enrollment_intensity_col`, but that's not the default. does this seem like an improvement, or a regression? fwiw Kelsey did something like this for her first pass on Lee College -- she derived an "average" intensity, rather than the "mode", and modified expected time limits accordingly; I didn't implement that back in the day because it was quite complicated and hard to reason around. this new implementation feels like a happy middle.


vishpillai123 (approved):
looks good!

kaylawilding:
Hi, I like the mode edit over the first enrollment and allowing that we can switch it easily back to first term enrollment. Should we worry about ties and set a default to part or full time?

bdewilde:
@kaylawilding Good question about ties — in case of multiple mode values (i.e. ties), the agg func just takes the first value. I believe there's even a test case that covers this. I think this is defensible :)

```python
def _mode_aggfunc(ser: pd.Series, *, dropna: bool = True) -> object:
    mode = ser.mode(dropna=dropna)
    return mode.iat[0] if not mode.empty else pd.NA
```

PR #77, opened 2025-02-17
[pdp] add selection + checkpoint subpackages

bdewilde:
## changes

- adds `checkpoints` subpackage with `pdp` module containing cleaner re-implementations of checkpointing functionality previously mixed into the `targets` code, as well as thorough unit tests
- adds `selection` subpackage with `pdp` module containing with cleaner re-implementations of student selection functionality previously mixed into `targets` code, as well as thorough unit tests
- adds util func to convert between year- and term-based intensity time limit inputs in various `targets`
- adds support for passing a wildcard `"*"` in for intensity, such that it applies the same time limits to all students regardless of enrollment intensity (this was requested by one of Noreen's schools)

**note:** fully half of this PR's LoC is unit tests. please feel free to look at them, or skim over and rest assured that this code is well-tested. :)

out-of-scope: removing equivalent/related functionality from its previous locations in `targets`. i'll clean that up in a separate pass.

## context

This is the next chunk of work in the decoupling of target calculations from other operations that are needed in contexts besides target calculation. It'll make it much easier to mix and match checkpoints, selections, and targets.

## questions

- I took a slightly different approach in the "checkpoints" functionality compared to what we had before. Basically, everything boils down to pre-filtering rows and using `DataFrame.groupby().nth()` in order to filter rows to a particular term per student. I added separate functions for some common checkpoints, purely for dev convenience and consistency, rather than over-parametrizing the lower-level functions. Does this make sense to you?


bdewilde:
an illustrative example:

```python
selected_students_by_attrs = selection.pdp.select_students_by_attributes(
    df_student_terms,
    student_id_cols=cfg.student_id_col,
    **cfg.preprocessing.target.student_criteria
)
if run_type == "train":
    # for train runs, we need to to filter by future (year-2) data availability
    # so that the computed targets are guaranteed to be accurate
    selected_students_by_year2_data = selection.pdp.select_students_by_second_year_data(
        df_student_terms,
        student_id_cols=cfg.student_id_col,
        cohort_id_col="cohort_id",
        term_id_col="term_id",
    )
    selected_students = selected_students_by_attrs.index.intersection(selected_students_by_year2_data.index)
else:
    # for predict runs, we want to filter by *the lack of* future data availability
    # i.e. first-year students who haven't yet gotten to year 2
    selected_students_by_year2_data = selection.pdp.select_students_by_second_year_data(
        df_student_terms,
        student_id_cols=cfg.student_id_col,
        cohort_id_col="cohort_id",
        term_id_col="term_id",
    )
    selected_students = selected_students_by_attrs.index.difference(selected_students_by_year2_data.index)

logging.info(
    "%s out of %s students selected overall",
    len(selected_students),
    df_student_terms[cfg.student_id_col].nunique()
)

df_ckpt = checkpoints.pdp.first_student_terms_within_cohort(
    df_student_terms,
    student_id_cols=cfg.student_id_col,
    sort_cols="term_rank",
    term_is_pre_cohort_col="term_is_pre_cohort",
)
df_features = pd.merge(df_ckpt, pd.Series(selected_students), how="inner", on=cfg.student_id_col)
```

src/student_success_tool/targets/pdp/shared.py
vishpillai123, line 1: what is in this file? Is it possible to change this name (from shared) to something more clear given that we have the same for the credits_earned and graduation.

src/student_success_tool/selection/pdp.py
vishpillai123, line 1: I think 'selection' terminology could be confused with feature_selection. Perhaps population_selection? Or something more clear?

src/student_success_tool/targets/pdp/shared.py
bdewilde: I'm confused :) This is the same file we've had before, it includes functionality shared by multiple target-specific modules

vishpillai123 (commented):
With these subpackage changes, would it be possible to also change the pdp templates to reflect these changes? Or open up a PR to reflect this in the notebooks. I'm fine with the changes but I think it can create avoidable issues when the changes aren't propagated in the notebooks.

src/student_success_tool/selection/pdp.py
bdewilde: I know, naming is hard when so many terms in the space are so overloaded. In the case of feature selection, it is nicely separated as `modeling.feature_selection`, so I was hoping there wouldn't be any possibility of confusion with a new `selection` subpackage. Mostly just trying to keep the subpackage names succinct and differentiated...

bdewilde:
> With these subpackage changes, would it be possible to also change the pdp templates to reflect these changes? Or open up a PR to reflect this in the notebooks. I'm fine with the changes but I think it can create avoidable issues when the changes aren't propagated in the notebooks.

I _will_ be doing this, but wanted to get this chain of updates merged before updating the templates just once, and other deadlines have been prioritized instead 😮‍💨

vishpillai123:
> > With these subpackage changes, would it be possible to also change the pdp templates to reflect these changes? Or open up a PR to reflect this in the notebooks. I'm fine with the changes but I think it can create avoidable issues when the changes aren't propagated in the notebooks.
> 
> I _will_ be doing this, but wanted to get this chain of updates merged before updating the templates just once, and other deadlines have been prioritized instead 😮‍💨

Okay, I think in the future, would it be possible to add everything in the same PR? Especially since these are more migration changes instead of new features that are added, it makes sense to ensure the templates always reflect the latest changes.

bdewilde:
@vishpillai123 Since these changes are additive and don't modify existing functionality -- this was intentional :) -- I think it's fine to not have them reflected in template notebooks. The template code should still work! And since all of this is going into the `develop` (i.e. under development) branch, I'm not pressed about the temporary mess and duplicative logic.

src/student_success_tool/selection/pdp.py
kaylawilding: Could we call it filtering? eligibility? eligible_students? I agree selection could get confusing

kaylawilding:
Heads up with this PR, @bdewilde I think you should look at the pipelines/tasks/pdp_data_preprocessing/pdp_data_preprocessing.ipynb script in this PR https://github.com/datakind/student-success-tool/pull/82. Mainly thinking of lines 106-108 where he uses the targets.shared.select_students_by_criteria function. Is this the one we should be using there? Maybe it's fine for now?

src/student_success_tool/selection/pdp.py
bdewilde: "selection" was the best word I could come up with! "filtering" is imo even more overloaded as a term, and slightly less descriptive of what the code actually does... Let me merge this in, but I will give it some thought and report back. Or we could chat about alternatives tomorrow in the weekly meeting (assuming we're having it)?

kaylawilding:
> > @vishpillai123 Since these changes are additive and don't modify existing functionality -- this was intentional :) -- I think it's fine to not have them reflected in template notebooks. The template code should still work! And since all of this is going into the `develop` (i.e. under development) branch, I'm not pressed about the temporary mess and duplicative logic.
> 
> I agree that the changes don't modify any existing functionality and technically we are still in `develop` so this isn't too concerning. I'm more trying to communicate how migrating without changing the template notebooks adds unnecessary work on our team in terms of digging through the repo and finding the new locations of subpackages. I'm fine with merging what you've done here, but I'm raising this concern also because I realized the inconvenience this created last week after I approved similar migration changes. My ask is more around if we can add the changes to template notebooks in the same PR next time we migrate subpackages. Adding @nm3224 and @kaylawilding for visibility here.

I agree that it would be key to update the template notebooks in the same PR if the code breaks the templates. Since this is additive and we are in the state we are in now, I think this is fine for now. I imagine (hope, fingers crossed) we don't have major changes going forward and updating the templates will be quick with PRs. I do agree that going forward we should try to update the template notebooks if possible (even if they dont break anything). Though, emphasis on the if possible if they don't break anything, because time constraints and priorities are real.

PR #78, opened 2025-02-21
Feature dv task

samroon2:
Adding code for a pipeline task for data validation. Validation task will help identify schema changes, and will enable drift detection etc. (sorry for the size, mostly uv updates!)

## changes
- Added new directory for pipelines and reusable tasks.
- Added code (that can be further extended) and simple working implementation along with readme and tests.
- Updated uv to include a separate group for data validation (made some other minor mods).

## context
[See Asana](https://app.asana.com/0/1208503162700153/1209242833795790).

## questions
- Happy to exclude uv mod. if this causes issues / isn't the intended way the team envisions dependency management (we'll also probably investigate the DB suggestion of managing task specific deps. at the cluster level).
- Leaving tests for this task separate from tests dir for now, will potentially update once we determine best strategy for dep. management and any potential changes to the CI workflow for tasks.

### DIFF
--- a/pipelines/tasks/data_validation/README.md
+++ b/pipelines/tasks/data_validation/README.md
@@ -15,7 +15,7 @@ The job requires the following parameters:
 
 * `--input_table_path`: Path to the input Delta table.
 * `--input_table_format`: Format of the input table (default: `delta`).
-* `--input_schema_path`: Path to the input reference schema pbtxt file.
+* `--input_schema_path`: Path to the schema pbtxt file.
 * `--output_artifact_path`: Path to write output artifacts (anomalies and statistics).
 * `--environment`: Environment to use for validation (`TRAINING` or `SERVING`).
 * `--fail_on_anomalies_true`: Fail the job if anomalies are found (default).
--- a/pipelines/tasks/data_validation/README.md
+++ b/pipelines/tasks/data_validation/README.md
@@ -1,0 +1,64 @@
+# Data Validation Task with TFDV
+
+This Databricks job uses [TensorFlow Data Validation (TFDV)](https://www.tensorflow.org/tfx/data_validation/get_started) to validate data against a reference schema and detect anomalies. It's designed for use in both training and inference pipelines.
+
+## How it Works
+
+1. **Loads a reference schema:** The job loads a schema.pbtxt file, this should be the expected schema determined during the training / experimentation phase of model development and maintained as need. It also assumes that there are two schema environments; TRAINING and SERVING, see [ref.](https://www.tensorflow.org/tfx/data_validation/get_started#schema_environments) for details.
+2. **Generates statistics:** It generates statistics from a Delta table or CSV file and persists them for further analysis (e.g. [checking skew and drift](https://www.tensorflow.org/tfx/data_validation/get_started#checking_data_skew_and_drift)).
+3. **Validates data:** It validates the data against the schema and generates a list of anomalies, optionally causing the task to fail upon detection.
+4. **Saves anomalies:** It saves the anomalies to a pbtxt file for review.
+
+## Job Parameters
+
+The job requires the following parameters:
+
+* `--input_table_path`: Path to the input Delta table or CSV file.
+* `--input_table_format`: Format of the input table (default: `delta`).
+* `--input_schema_path`: Path to the schema pbtxt file.
+* `--output_artifact_path`: Path to write output artifacts (anomalies and statistics).
+* `--environment`: Environment to use for validation (`TRAINING` or `SERVING`).
+* `--fail_on_anomalies_true`: Fail the job if anomalies are found (default).
+* `--fail_on_anomalies_false`: Do not fail the job if anomalies are found.
+
+**NOTE:** `--fail_on_anomalies_true` and `fail_on_anomalies_false` are a [mutually exclusive group](https://docs.python.org/3/library/argparse.html#mutual-exclusion), only one per run should be passed to the task.
+
+## How to Run
+
+1. **Create a Databricks Job:** In your Databricks workspace, create a new job.
+2. **Add a Task:** Add a Python task to the job.
+3. **Configure the Task:**
+    * **Python File:** Upload the Python script to DBFS and provide the path in the task configuration (git repo optional as well).
+    * **Parameters:** Provide the required parameters as described above.
+4. **Run the Job:** Run the job to validate your data.
+
+**NOTE:** You can also easily integrate this task with other pipelines / jobs / tasks as well!
+
+## Example
+
+To validate data in a Delta table against a schema and fail the job if anomalies are found:
+```
+[
+"--input_table_path", "/path/to/your/delta/table",
+"--input_schema_path", "/path/to/your/schema.pbtxt",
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -5,17 +5,14 @@ description = "School-agnostic lib for implementing Student Success Tool workflo
...
-    "databricks-connect~=16.0",
+    "databricks-connect",
     "databricks-sdk~=0.38.0",
     "faker>=20.1",
     "ipywidgets~=8.1",
     "matplotlib~=3.7",
     "missingno~=0.5",
     "mlflow~=2.13",
-    "numpy~=1.26",
-    "pandas~=2.1",
```
bdewilde, line 16: Why have you removed `numpy` and moved `pandas`? We use them all over the place, and need to specify+constrain their versions -- especially important since `numpy` v2 breaks things.

pyproject.toml
```diff
@@ -5,17 +5,14 @@ description = "School-agnostic lib for implementing Student Success Tool workflo
 readme = "README.md"
 requires-python = ">=3.10,<3.13"
 dependencies = [
-    "databricks-connect~=16.0",
+    "databricks-connect",
```
bdewilde, line 8: Why are you removing this version constraint?

pyproject.toml
```diff
@@ -36,13 +33,24 @@ dev = [
...
 build-backend = "hatchling.build"
 
 [tool.uv]
 default-groups = ["dev"]
+conflicts = [
+    [
+      { group = "group1" },
+      { group = "data-validation" }
+    ],
+]
```
bdewilde, line 53: Can you explain this? I'm not familiar with the syntax / feature / causes.

pipelines/tasks/data_validation/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+The job requires the following parameters:
+
+* `--input_table_path`: Path to the input Delta table.
+* `--input_table_format`: Format of the input table (default: `delta`).
+* `--input_schema_path`: Path to the input reference schema pbtxt file.
+* `--output_artifact_path`: Path to write output artifacts (anomalies and statistics).
+* `--environment`: Environment to use for validation (`TRAINING` or `SERVING`).
+* `--fail_on_anomalies_true`: Fail the job if anomalies are found (default).
+* `--fail_on_anomalies_false`: Do not fail the job if anomalies are found.
```
bdewilde, line 22: Can this be specified as a single flag? e.g. `parser.add_argument("--fail-on-anomalies", action="store_true", default=False)`

bdewilde (commented):
This is just a partial review, since I'm not sure I understand how this functionality will be used in practice, and I'm not familiar with tensorflow data validation. Call me old school, but it feels like we could implement this sort of thing "in-house"? Mostly I'm noticing that this is [a heavy dependency](https://github.com/tensorflow/data-validation/blob/master/setup.py#L176-L208) with hard version incompatibilities on core libs, such as `pandas` -- a strong yellow flag.

vishpillai123:
hey @samroon2 - I agree with Burton we can't remove `numpy` and `pandas` versioning here as it's quite critical to the SST package. I'm assuming this was due to dependency conflicts with `tfdv`, so I'm curious if we could create a similar setup with `tfdv` using `pandera`, since we use it almost everywhere already with schema checks. 

What other benefits does `tfdv` offer? Is it because we can also check for data/model drift? I think I remember you saying that. If I understand correctly, it looks like for this implementation, we're only checking for schema anomalies. So, I think `pandera` could cover our bases here.

samroon2:
@vishpillai123 @bdewilde thanks for taking a look! For the dependency piece, as I mentioned in the PR desc. it's not super clear what the overall management strategy is for this repo, and what exactly the “sst package” should encompass… It kind of seems like it’s moving toward being web app and DS + ML deps. to be managed in the same uv env. which might not be the best strategy imo (and seems like it might be a path for future tech. debt). ML systems alone can get pretty complex and bps are typically to containerize each component / tasks to ensure environment isolation, consistency across envs., easier management, portability, reproducibility and scale. For now, I’m going to roll back the uv changes and go with the DB suggestions for isolating deps. between tasks, happy to chat more about this though (also, if you’re interested in learning more about the uv groups you can check it out here: https://docs.astral.sh/uv/concepts/projects/config/#conflicting-dependencies).

As for tfdv I think we’ve discussed this a few times and have described the usage as intended for this particular task in the design doc we shared / y’all reviewed (tldr; schema validation (can be any schema not just the raw pdp), anomaly detection, drift + skew detection, plus easy viz solutions when manually reviewing). I personally wouldn’t recommend reimplementing something in-house to achieve a similar objective, but I could see if there was a DB managed alternative solution that offered similar functionality that might be a logical substitution for future evaluation (although tfdv would offer a more portable solution). I’d also suggest maybe checking tfdv out as a sub for pandera, it seems like it could potentially offer a few more features (and would really help expedite adding new schemas, particularly if they are highly custom), though I’m not super familiar with pandera so y’all might have a better view on that!

pyproject.toml
samroon2: You can check it out [here in the uv docs](https://docs.astral.sh/uv/concepts/projects/config/#conflicting-dependencies).

bdewilde:
> it’s moving toward being web app and DS + ML deps. to be managed in the same uv env

This is news to me! I definitely was not aware that the web app is moving in here. That's _fine_, but I think we can and should keep those envs separate. Would uv workspaces help here? https://docs.astral.sh/uv/concepts/projects/workspaces/

pipelines/tasks/data_validation/README.md
samroon2: explicit > implicit in this scenario imo.

bdewilde: I strongly disagree, boolean flags have been common in CLI commands for decades.

vishpillai123:
> > it’s moving toward being web app and DS + ML deps. to be managed in the same uv env
> 
> This is news to me! I definitely was not aware that the web app is moving in here. That's _fine_, but I think we can and should keep those envs separate. Would uv workspaces help here? https://docs.astral.sh/uv/concepts/projects/workspaces/
> 
> > I’d also suggest maybe checking tfdv out as a sub for pandera, it seems like it could potentially offer a few more features
> 
> Noted, but I don't think we have bandwidth for another major technical migration at this time. Gotta pick our battles!

I also wasn't aware that web app is moving here. Have we aligned on that already? I think it makes more sense to separate, but I'm not entirely sure how the web app works and I haven't been part of those discussions.

samroon2:
Yeah I agree it doesn't really make sense, all the web app work is currently on a different branch (fellows-experimental), we can bring this up for a topic to discuss tomorrow with the broader group. As for this PR, the changes to uv have been rolled back, are any outstanding issues?

vishpillai123:
> Yeah I agree it doesn't really make sense, all the web app work is currently on a different branch (fellows-experimental), we can bring this up for a topic to discuss tomorrow with the broader group. As for this PR, the changes to uv have been rolled back, are there any outstanding issues?

Okay sounds great!! For now, I don't have concerns on my end. Waiting on Burton's approval here as well before we merge.

vishpillai123 (approved):

PR #83, opened 2025-02-23
[pdp] fix bugs and devx rough spots

bdewilde:
## changes

it's a grab-bag!

- bumps `uv` in CI to latest version, makes style check outputs useful in case of error, and bumps `ruff` from v0.8 => v0.9
- generalizes logic to create a certain class of human-readable feature names, and adds a minor new feature to leverage it
- fixes a type check bug in a model eval plot
- adds filtering of constant-value (non-numeric) columns during the low-variance feature selection; this is the equivalent of `threshold=0.0` for numeric columns, so makes sense here no matter the threshold
- adds functions for naming models, getting their uris, and registering them via mlflow
- improves arg input when computing dataset splits
- fixes scary bug in future data masking function, such that features defined at the year level that are in the same year as the current year of enrollment get set to null; this is important because, for example, previously first-term rows could have first-year features populated, which is usually data leakage from the future; this isn't always the case, but we have no good way to know for sure, so better to be safe

## context

working my way through a minor USCB pipeline update, and hitting snags

## questions

- have you run into any other rough edges that you'd like to see fixed here?


bdewilde:
hey hey, i know you're busy, so i'm going to merge this in w/o review 😇  most of these changes are minor additions that make development nicer. the only thing that i definitely want your feedback on is 

> fixes scary bug in future data masking function, such that features defined at the year level that are in the same year as the current year of enrollment get set to null; this is important because, for example, previously first-term rows could have first-year features populated, which is usually data leakage from the future; this isn't always the case, but we have no good way to know for sure, so better to be safe

i can always change this after the fact, if you feel strongly in the other direction!

PR #85, opened 2025-02-26
Add utils library needed for publishing output to GCP buckets

crysg:
## changes
This is a library needed by the inference task in the Databricks that will copy over the output file (e.g. inference output csv) to the bucket in GCP such that the webapp will be able to access it.


crysg:
Oops sorry, fixing some typos that I introduced after cleaning up the file

vishpillai123 (changes requested):
hi @crysg, the structure of this file & the functions look good to me! The location I think may belong more [here](https://github.com/datakind/student-success-tool/tree/develop/pipelines/inference/pdp_inference) right? Not sure how this utils file is being used but I assumed it was in the pipeline.

src/student_success_tool/gcsutils.py
```diff
@@ -0,0 +1,71 @@
+"""
+Utility functions that interact with Google Cloud Storage.
+For instance, copying files from Databricks volumes into GCP buckets.
+"""
+
+from google.cloud import storage
+from google.cloud.storage.bucket import Bucket
```
bdewilde, line 7: I don't see this package listed as a dependency in `pyproject.toml` -- where's it coming from?

src/student_success_tool/gcsutils.py
```diff
@@ -0,0 +1,71 @@
...
+    """Save file from databricks volume to a GCP bucket path."""
+    blob = bucket.blob(dest_bucket_pathname)
+    if blob.exists():
+        raise ValueError(dest_bucket_pathname + ": File already exists in bucket.")
+    blob.upload_from_filename(src_volume_filepath)
+
+
+def get_filename_from_file_path(file_path: str) -> str:
+    """Get the filename only of a given filepath, e.g. for home/directory/file.csv, return just file.csv."""
+    return file_path.split("/")[-1]
```
bdewilde, line 23: iiuc, you should just use the stdlib implementation: https://docs.python.org/3/library/os.path.html#os.path.basename

bdewilde:
looks like you've added quite a bit more than just a utils module! what happened?

crysg:
> looks like you've added quite a bit more than just a utils module! what happened?

crysg:
Oh oops, I'm trying to pick up the recent changes from develop, but I don't know what happened...

PR #86, opened 2025-02-27
Add utils files used by pipeline in utils directory

crysg:
Adding some utils directories that are interacting with GCS and SMTP into the pipeline utils. (We expect these will only be used by the pipeline tasks).

## changes
Adding two utils files: emails.py and gcsutils.py. emails.py creates SMTP server and sends an email. gcsutils.py copies files from databricks volume to GCP bucket using code from GCP documentation.

## context
Resolved the comments in the other PR where I messed up the merge with develop.


crysg:
> Hi @crysg ive botched merges using databricks in a similar way. It’s super strange how that happens. I also usually ditch the PR and open a new one like you have lol.
> 
> Thanks for moving the utils file. I’m seeing a 3 other files now that have been changed though. Was this intentional?

Ah good to know! I need to read up on how to use Github...

Yes that's intentional! I added the emails utils earlier, and that should belong with the gcutils here. The other files are uv package management changes to include google cloud storage python library as a dependency. This dependency is required because the gcsutil.py I'm adding uses GCS library calls to handle GCP bucket changes (file copies etc.)

pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
     "statsmodels~=0.14",
     "shap~=0.46.0",
     "tomli~=2.0; python_version<'3.11'",
+    "google-cloud>=0.34.0",
```
bdewilde, line 26: nit: slot this into the dependency list in alphabetical order

pipelines/tasks/utils/gcsutils.py
```diff
@@ -0,0 +1,67 @@
...
+    """Save file from databricks volume to a GCP bucket path."""
+    blob = bucket.blob(dest_bucket_pathname)
+    if blob.exists():
+        raise ValueError(dest_bucket_pathname + ": File already exists in bucket.")
+    blob.upload_from_filename(src_volume_filepath)
+
+
+def publish_inference_output_files(
+    db_workspace: str,
+    institution_name: str,
```
bdewilde, line 24: is this the institution "name" or "id"? based on how this variable is used, i _think_ it's actually the id.

pipelines/tasks/utils/gcsutils.py
```diff
@@ -0,0 +1,67 @@
...
+    external_bucket_name: str,
+    sst_job_id: str,
+    approved: bool,
+) -> str:
+    """
+    Publish output files to bucket, with folder determined based on approved parameter,
+    and return the status of the job as a string.
+    """
+    # Destination: The output directory in the bucket that the files get moved to.
+    bucket_directory = f"{'approved' if approved else 'unapproved'}/{sst_job_id}"
```
bdewilde, line 34: nit: i think this is technically correct, but python has a cleaner syntax for conditional assignment:

```suggestion
    bucket_directory = "approved" if approved else f"unapproved/{sst_job_id}"
```

pipelines/tasks/utils/gcsutils.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+
+
+def save_file(
+    bucket: Bucket, src_volume_filepath: str, dest_bucket_pathname: str
+) -> None:
+    """Save file from databricks volume to a GCP bucket path."""
+    blob = bucket.blob(dest_bucket_pathname)
+    if blob.exists():
+        raise ValueError(dest_bucket_pathname + ": File already exists in bucket.")
```
bdewilde, line 18: is there ever a use case where you'd want to overwrite the contents of a bucket? if so, maybe a `force: bool = False` param and logic to handle it would be nice?

crysg: As of my current understanding of constraints/requirements/the feature request of the publish etc, no. But easy to add a param here in the future if that requirement ever comes up.

pipelines/tasks/utils/gcsutils.py
crysg: The webapp has a concept of both an institution ID (a UUID generated by the database to obfuscate institution names and make sure they are uniquely identified) and institution name, this parameter is analogous to the institution name (though "databricksified") as understood by the webapp, NOT the institution ID. However, I took a look at some of your templates and it looks like it maps to the institution ID in those templates.

Since this is a utility function mainly used by the pipeline which is integrated with the webapp, it'll need to be consistent with the webapp to avoid confusion. People should not be manually enter the parameters here, the inference pipeline will be triggered from the webapp.

pyproject.toml
crysg: Done.

vishpillai123 (approved):
Moving forward based on offline conversation. Crystal to add documentation on "inst_id" definition to README, since there's some inconsistency with what google fellows/DK are calling "inst_id" and "inst_name"

pipelines/tasks/utils/gcsutils.py
bdewilde: oops, sure enough! i misread your code. imo that underscores my point: doing this in an f-string is straight-up harder to read, even though it's technically allowed :) too late now, and no big deal

PR #88, opened 2025-03-02
[pdp] handle new raw cols + add new features

bdewilde:
## changes

- adds functionality to add empty columns raw course/cohort data in case those columns are missing; this is a straightforward way of handling new columns added by pdp w/o requiring tons of checks in features code
- correctly sets new year-to raw cohort fields to null in case of 0 values
- adds new "term program of study area" feature, based on new "term program of study" raw column, plus a derived "number of courses in term program of study area" feature
- adds course level mean to the set of features for which "term diff" values are computed
- adds new features to mapping table

## context

PDP added a handful of new columns to the raw data; the logic here properly integrates them into the existing feature set, and adds a couple features on top.

## questions

- any concerns about adding empty columns in case the raw data is missing them? presumably this won't cause any trouble for already trained models (owing to their column selectors), and in case of all-null values, presumably feature selection would automatically drop them anyway...


bdewilde:
howdy @vishpillai123 @kaylawilding , friendly nudge to not forget about this PR 🙇

vishpillai123 (approved):

PR #90, opened 2025-03-04
[pdp] improve handling for unique student identifiers

bdewilde:
## changes

- bumps `pandera` dep from v0.21 to v0.23
- replaces `"study_id"` column renaming hack with consistent renaming of `"student_guid"` / `"study_id"` cols to `"student_id"`, within the raw data schemas themselves, so that both forms of raw PDP data are handled internally in the same way
- updates default student id col(s) arg values to `"student_id"` throughout code base

## context

for _reasons_ PDP sends us data in two forms, where one uses "student_guid" as a unique student identifier while the other form uses "study_id"; they serve the same purpose, and are never used in a dataset at the same time. rather than having to switch back-and-forth between student id col names -- a big hassle -- these changes use a consistent, obvious name under the hood.

## questions

- any school-specific pipelines pointed at the `develop` branch here will have to update their project configs, i.e. `student_id_col = "student_id"`. any concerns about this?


bdewilde:
I saw your comments on other schema PR, @vishpillai123 , but friendly reminder that this one has to be merged first. :)

nm3224:
Looks all good to me! Maybe we should send out a slack or something letting people know they'll need to change student_id in their configs but I think it's an easy change if anyone is still using the develop branch!

nm3224 (approved):

PR #91, opened 2025-03-05
Move utils

crysg:
We are having a terrible time in Databricks trying to make the imports from outside student-success-tool/src/ work in Databricks via a non-hacky way.

## changes
Move the utils functions into a location readable by Databricks pipelines

Also add the chart util file.


vishpillai123:
@bdewilde the fellows asked this in the meeting today. I totally understand (and can relate) to the issues with dependencies they are having. Though, I wasn't sure how you would feel about this? Can you take a look?

bdewilde (changes requested):
It looks like you moved these utility modules and also added some type annotations. The annotations seem fine (but I have very little context for them). The new location of the files does give me pause: You've created a new top-level directory in the repo, which is outside the main Python package. How/where is this being imported? How does this move fix things??

bdewilde (approved):
Hey @crysg , I just realized that the full file path was obscured by GitHub, so I mis-read that you did indeed move this into a directory within `src/student_success_tool`, rather than the top-level directory I read. Sorry, that was UX-facilitated user error 😇 

I think this is fine. If you'd prefer these nb utils modules to be in their old home, you might try path-hacking? This is a thing Databricks recommends, though afaik it is strongly discouraged in general.

```python
if "path/to/utils/dir" not in sys.path:
    sys.path.insert(1, "path/to/utils/dir")

from utils import [stuff]
```

crysg:
No worries, @bdewilde! And thanks for the quick response.

Yes we tried path hacking, but came to the same conclusion you did which is it's much cleaner to simply move our utils library somewhere that the python files can read.

PR #92, opened 2025-03-05
Created a utility function to generate a shap beeswarm plot

jamalc:
## changes

This change creates a new file in pipeline utils to plot a beeswarm chart.
## context

The summary_plot method of generating charts for shap data is deprecated.


crysg:
I think you have to run the formatter but otherwise lgtm!

vishpillai123 (approved):
not familiar with this type of plot, so curious to see it!

PR #94, opened 2025-03-06
Using the new Shap explanation api

pemujo:
Updated the pdp_model_inference task to use the ["New Shap Explanation API"](https://shap.readthedocs.io/en/latest/example_notebooks/api_examples/migrating-to-new-api.html) and plotted using beeswarm function. 

The changes were required because some methods used with the old API are getting deprecated and Shap documentation pointed to the new API.

## Changes
**1. Removed the custom code to create shap values**

```
shap_schema = StructType(
    [StructField(cfg.student_id_col, StringType(), nullable=False)]
    + [StructField(col, FloatType(), nullable=False) for col in model_feature_names]
)

df_shap_values = (
    spark_session.createDataFrame(
        df_processed_dataset.reindex(
            columns=model_feature_names + [cfg.student_id_col]
        )
    )
    .repartition(spark_session.sparkContext.defaultParallelism)
    .mapInPandas(
        ft.partial(
            inference.calculate_shap_values_spark_udf,
            student_id_col=cfg.student_id_col,
            model_features=model_feature_names,
            explainer=explainer,
            mode=train_mode,
        ),
        schema=shap_schema,
    )
    .toPandas()
    .set_index(cfg.student_id_col)
    .reindex(df_processed_dataset[cfg.student_id_col])
    .reset_index(drop=False)
)
```

**Replaced with the new explanation API**
```
    # Calculate Shap values using the new Shap Explanation API
    df_shap_values = explainer(df_processed_dataset[model_feature_names])
```

**2. Removed the plot function**
```
shap.summary_plot(
      df_shap_values.loc[:, model_feature_names].to_numpy(),
      df_serving_dataset.loc[:, model_feature_names],
      class_names=loaded_model.classes_,
      max_display=20,
      show=False,
  )
```

**replaced with beeswarm utils function**
```
from student_success_tool.pipeline_utils.plot import plot_shap_beeswarm
# Plot Shap values using the beeswarm plot function 
 shap_fig = `plot_shap_beeswarm(df_shap_values)`
```
3. Removed the unsed dependencies.

## Context
Per documentation:
"The new-style plotting functions like shap.plot.bar and shap.plots.beeswarm accept these Explanation objections rather than numpy arrays." 
 

    

## Questions/Inputs
The performance was similar to the custom spark function with synthetic datasets, but it should be tested with bigger datasets


pipelines/tasks/pdp_model_inference/pdp_model_inference.ipynb
```diff
@@ -250,63 +237,24 @@
...
-    "            ),\n",
-    "            schema=shap_schema,\n",
-    "        )\n",
-    "        .toPandas()\n",
-    "        .set_index(cfg.student_id_col)\n",
-    "        .reindex(df_processed_dataset[cfg.student_id_col])\n",
-    "        .reset_index(drop=False)\n",
-    "    )\n",
+    "    # Calculate Shap values using the new Shap Explanation API\n",
+    "    df_shap_values = explainer(df_processed_dataset[model_feature_names])\n",
```
vishpillai123, line 241: I'm not seeing where you imported the explainer function here. Also not sure how performance will be for thousands of students, so this may need to be parallelized.

pipelines/tasks/pdp_model_inference/pdp_model_inference.ipynb
```diff
@@ -250,63 +237,24 @@
...
-    "    #     mlflow.log_figure(\n",
-    "    #         shap_fig,\n",
-    "    #         f\"shap_summary_dataset_name_dataset_\"\n",
-    "    #         f\"{df_ref.shape}_ref_rows.png\",\n",
-    "    #     )\n",
+    "    # Plot Shap values using the beeswarm plot function\n",
+    "    shap_fig = plot_shap_beeswarm(df_shap_values)\n",
     "\n",
     "    # --- Feature Selection for Display ---\n",
     "    result = inference.select_top_features_for_display(\n",
```
vishpillai123, line 248: If we are using the values directly here, can we validating that results here make sense (preferrably in a unit test)? Since we're using this across schools, unit tests can help prevent future bugs.

For example, for each student, does it follow that top X features have descending SHAP values (1>2>3>4>5)?

pipelines/tasks/pdp_model_inference/pdp_model_inference.ipynb
```diff
@@ -250,63 +237,24 @@
...
-    "\n",
-    "    # # --- Log SHAP Plot to MLflow ---\n",
-    "    # with mlflow.start_run(run_id=model_run_id) as run:\n",
-    "    #     mlflow.log_figure(\n",
-    "    #         shap_fig,\n",
-    "    #         f\"shap_summary_dataset_name_dataset_\"\n",
-    "    #         f\"{df_ref.shape}_ref_rows.png\",\n",
-    "    #     )\n",
+    "    # Plot Shap values using the beeswarm plot function\n",
+    "    shap_fig = plot_shap_beeswarm(df_shap_values)\n",
```
vishpillai123, line 245: How are we visualizing without the names of the model features?

pipelines/tasks/pdp_model_inference/pdp_model_inference.ipynb
pemujo: hey @vishpillai123 df_shap_values is actually a shap.explanation object that includes the values and features names. 

This is a snippet from the [documentation](https://shap.readthedocs.io/en/latest/example_notebooks/api_examples/migrating-to-new-api.html):
**"The shap.Explanation object is a much richer representation that includes the shap values (accessible with the .values attribute) as well as supporting contextual information such as the background dataset and the feature names."**

I think it would be good to change the name to reflect that. df_shap_values could be renamed to shap_explanations or something that properly reflects the difference.

pipelines/tasks/pdp_model_inference/pdp_model_inference.ipynb
pemujo: We did not change that part. We still used the explainer as it was before:

```    
explainer = shap.explainers.KernelExplainer(
        ft.partial(
            predict_proba,
            model=loaded_model,
            feature_names=model_feature_names,
            pos_label=cfg.pos_label,
        ),
        df_ref,
        link="identity",
    )
```

https://github.com/datakind/student-success-tool/blob/6a2bd980b64e0c8339b60bc54a2677bfef18d979/pipelines/tasks/pdp_model_inference/pdp_model_inference.ipynb#L229

Completely agree about performance. This will need to be tested and you could refactor to parallelize the function if needed

pipelines/tasks/pdp_model_inference/pdp_model_inference.ipynb
pemujo: Unit test would be ideal. That is one of the main reasons to switch to .py instead of notebooks

vishpillai123: Sure, but you can build unit tests and have notebooks. Our best practice at DK is to separate functions into .py files and then import them into notebooks. That way you can still create unit tests. 

I would say almost all of the functionality (if not all, Burton is very detailed) we have here in the public repo has unit tests ([reference](https://github.com/datakind/student-success-tool/tree/develop/tests)).

pipelines/tasks/pdp_model_inference/pdp_model_inference.ipynb
vishpillai123: ah interesting, so you put it directly into the explainer! Gotcha, that's pretty nice.

pipelines/tasks/pdp_model_inference/pdp_model_inference.ipynb
vishpillai123: If we don't want to go the unit test route, we can at the least have an assertion statement here to validate the results.

pipelines/tasks/pdp_model_inference/pdp_model_inference.ipynb
kaylawilding: They are part of the Explanation object and beeswarm can use that directly. 

_The shap.Explanation object is a much richer representation that includes the shap values (accessible with the .values attribute) as well as supporting contextual information such as the background dataset and the feature names.

The new-style plotting functions like shap.plot.bar and shap.plots.beeswarm accept these Explanation objections rather than numpy arrays._

What I do think we should do, small detail, to be equivalent to what we had before for plotting is to set the option max_display to 20 (shap.plots.beeswarm(shap_values, max_display=20).

kaylawilding (approved):
I have one small change about setting the max_display to 20 but I don't think its a blocker, so approving. Read below for more detail, but Im good with the current unit tests. Thanks for the updates here!

pipelines/tasks/pdp_model_inference/pdp_model_inference.ipynb
vishpillai123: I agree that unit tests on the function itself is silly, that’s not what I’m asking for. I would like to validate that the SHAP values make sense. 

I know it seems like the unit tests are silly as is. I think the correct shape, numeric, and nonempty checks are simple checks, but still important to have in my opinion. The student ID check is absolutely necessary (reminder that this used to be a bug in our previous SHAP values calc). I would advise to not delete the unit tests.

PR #95, opened 2025-03-06
[pdp] handle raw data more flexibly in base schemas

bdewilde:
## changes

- changes many non-feature fields from strict categoricals to flexible but standardized strings
- for many categorical fields, standardize their values and set their categories in a flexible, consistent manner; it makes the schemas less readable but more robust to weird data
- removes obsolete pre-schema data transformation logic from dataio functions; the schemas handle it now, which is as it should be

## context

goog folks want the base schemas to "just work" for more schools, which is in tension with the safety of the data especially wrt feature engineering; these changes are, hopefully, a workable middle ground

note that this PR builds on PR #90 -- and meaningfully so. we'll need that PR merged before this one.

## questions

- have i missed any fields in need of similar treatment? note that i didn't do this for _every_ field, if only bc we've never had trouble with them.


src/student_success_tool/schemas/pdp/raw_cohort.py
```diff
@@ -88,48 +52,15 @@ class RawPDPCohortDataSchema(pda.DataFrameModel):
...
-                "Transfer-In Full-Time",
-                "Transfer-In Part-Time",
-                "Transfer-In Unknown",
-                "Re-admit Full-Time",
-                "Re-admit Part-Time",
-                "Re-admit Unknown",
-            ]
-        },
-    )
-    credential_type_sought_year_1: pt.Series[pd.CategoricalDtype] = pda.Field(
```
vishpillai123, line 114: I think there's definitely less risk with not strictly setting the categories for bias columns. But, for `credential_type_sought_year_1`, don't we use this for later preprocessing/filtering/outcome variable def? Seems riskier.

Could we keep this column the same as it was (same categories)?

src/student_success_tool/schemas/pdp/raw_cohort.py
```diff
@@ -313,41 +244,67 @@ def rename_student_id_col(cls, df):
...
     def set_dual_and_summer_enrollment_categories(cls, series):
-        return series.cat.set_categories(["DE", "SE", "DS"])
+        return _strip_upper_strings_to_cats(series).cat.set_categories(
+            ["DE", "SE", "DS"]
+        )
 
     @pda.parser("pell_status_first_year")
     def set_pell_status_first_year_categories(cls, series):
-        return series.cat.set_categories(["Y", "N"])
+        return _strip_upper_strings_to_cats(series).cat.set_categories(["Y", "N"])
```
vishpillai123, line 287: on board with white space stripping, nice!

src/student_success_tool/schemas/pdp/raw_cohort.py
```diff
@@ -72,11 +40,7 @@ class RawPDPCohortDataSchema(pda.DataFrameModel):
     institution_id: pt.Series["string"]
     cohort: pt.Series["string"]
     cohort_term: pt.Series[pd.CategoricalDtype] = TermField()
-    student_age: pt.Series[pd.CategoricalDtype] = StudentAgeField()
-    enrollment_type: pt.Series[pd.CategoricalDtype] = pda.Field(
```
vishpillai123, line 76: Enrollment type also seems like it could break some of our target variables/filtering?

src/student_success_tool/schemas/pdp/raw_cohort.py
crysg: drive-by comment: another generally applied standardization function is lower() -- do you ever see cases where the case is variable? In which case it might make sense to apply that + whitespace stripping everywhere you can

src/student_success_tool/schemas/pdp/raw_cohort.py
bdewilde: Enrollment type is the same as it was, except we now set categories more flexibly -- see `set_enrollment_type_categories` below. Now, we automatically strip/uppercase the input strings and then set the cats such that values falling outside those cats are set to null rather than raise an error.

src/student_success_tool/schemas/pdp/raw_cohort.py
bdewilde: This columns has seen multiple variations across schools, and it's not used for _features_, just selecting students. Since that selection process is manual, I figured this would be safe to loosen up a bit.

src/student_success_tool/schemas/pdp/raw_cohort.py
bdewilde: This function does `upper()` instead, but same idea.

crysg: ok nice!

src/student_success_tool/schemas/pdp/raw_cohort.py
vishpillai123: Ok sounds good!

src/student_success_tool/schemas/pdp/raw_cohort.py
bdewilde: Yeah, that's a concern -- however, it would mean that zero students would be selected, and we'd notice pretty quickly that there was an issue, just at a slightly later point in the pipeline :)

vishpillai123: May be zero students or just less than what it should be. For example, if only one of the categories changes and our filtering has 2-3 categories etc, then it may not be immediately obvious.

vishpillai123 (approved):

src/student_success_tool/schemas/pdp/raw_cohort.py
bdewilde: Thanks @vishpillai123 , we can revisit this later. I understand your concerns, just trying to find a workable compromise.

PR #96, opened 2025-03-07
[inference validation] ensuring top features & shap values are descending

vishpillai123:
## changes
- Added validation step to check whether SHAP values are actually ranking correctly in importance.
- Added assertion that SHAP values are in strictly descending order for each student
- Updated pytest to have a check for when SHAP values are actually not ranking in order.
- Added rounding after the check. 

## context
- Fellows are adding a new shap calculation and I want to make sure the SHAP calculation makes sense.
- We need validation either way in our final output and it makes sense to me to embed it directly in our top features function since they are using the SHAP library function directly.
- If SHAP values are not actually ranking in correct order, it may suggest the SHAP calculation is not correct or perhaps processing is not correct.
- Since we're comparing floats it would be very unlikely in practice that we would run into an issue where two features have the exact same importance. 

## questions
- Let me know if you have any questions or concerns.


crysg:
> Since we're comparing floats it would be very unlikely in practice that we would run into an issue where two features have the exact same importance.


Just curious about this as I'm very new to shap -- would it be a strict _error_ if two shap values happened to have the same value? What happens if one feature is really just a reinterpretation of another feature or a transformation of another feature? Could they have the same value then? And if so, would that be considered an issue on the feature engineering side?

I wonder if it might be better for this to be a warning instead of error if it's not a strict error to have two of the same values. But will defer to the ML folks :)

crysg (approved):

vishpillai123:
> > Since we're comparing floats it would be very unlikely in practice that we would run into an issue where two features have the exact same importance.
> 
> Just curious about this as I'm very new to shap -- would it be a strict _error_ if two shap values happened to have the same value? What happens if one feature is really just a reinterpretation of another feature or a transformation of another feature? Could they have the same value then? And if so, would that be considered an issue on the feature engineering side?
> 
> I wonder if it might be better for this to be a warning instead of error if it's not a strict error to have two of the same values. But will defer to the ML folks :)

So that would be an example of covariance among two features in our model, which our feature selection attempts to account for, though covariance is a spectrum and it may be inevitable that some our features may have some relationship to eachother. With that said, as far as your question, I think if they have the same value I would be concerned because it's just very unlikely as an output from SHAP. We're talking like upwards to 10+ floating point numbers.

src/student_success_tool/modeling/inference.py
```diff
@@ -80,11 +80,26 @@ def select_top_features_for_display(
...
+    assert all(
+        all(
+            abs(top_features_df.loc[i, f"Feature_{j}_Importance"])
+            > abs(top_features_df.loc[i, f"Feature_{j + 1}_Importance"])
+            for j in range(1, n_features)
+        )
+        for i in range(len(top_features_df))
+    ), (
+        f"Final output has invalid SHAP values across top {n_features} ranked features for one or more students."
+    )
```
bdewilde, line 98: this feels like it belongs in a test suite -- and maybe not even this function's tests! what's it doing here?

src/student_success_tool/modeling/inference.py
```diff
@@ -80,11 +80,26 @@ def select_top_features_for_display(
...
+    # Round after assertion check
+    importance_cols = [f"Feature_{i}_Importance" for i in range(1, n_features + 1)]
+    top_features_df[importance_cols] = top_features_df[importance_cols].round(2)
```
bdewilde, line 101:
```suggestion
    top_features_df = top_features_df.assign(
        **{
            f"Feature_{i}_Importance": lambda df: df[f"Feature_{i}_Importance"].round(2)
            for i in range(1, n_features + 1)]
    )
```

tests/modeling/test_inference.py
```diff
@@ -83,20 +87,13 @@ def explainer():
...
-                {
-                    "Student ID": [1, 2, 3],
-                    "Support Score": [0.9, 0.1, 0.5],
-                    "Feature_1_Name": ["x1", "x2", "x4"],
-                    "Feature_1_Value": ["val1", "False", "3"],
-                    "Feature_1_Importance": [1.0, -1.0, 0.75],
-                }
-            ),
+            pd.DataFrame(),
+            True,  # Assertion error expected because SHAP values are out of order
```
bdewilde, line 96: it's hard to see what's going on here bc the git diff is weird -- did you add a new test case, or modify an existing one?

vishpillai123: Added a new test case!

src/student_success_tool/modeling/inference.py
vishpillai123: Yeah good question - Google fellows are using the SHAP explainer directly. We don’t have validation in the same way as before since we previously had unit tests that validated SHAP calculation. Anyways, it felt this would be a good place to add the assertion that SHAP values are descending because otherwise we don’t have validation with what they’re doing directly.

vishpillai123: The fellows aren’t using any of our functions so what test would we add it to? That’s why I decided to add it here.

bdewilde: Got it. My two cents: The fellows should use our functions rather than creating their own :) and if that meant updating the functionality to shap's current preferred API, all the better! But this is probably fine for now...

PR #98, opened 2025-03-08
[pdp] use new-style shap api

bdewilde:
## changes

uses new "Explanation" API under the hood when calculating SHAP values

## context

related: https://github.com/datakind/student-success-tool/pull/96

## questions

- is it possible to return `Explanation` objects instead of constructing DataFrames, such that we could run this in parallel/distributed fashion and then... merge?... the explanations? specifically talking about this mess:

```python
shap_schema = StructType(
    [StructField(cfg.student_id_col, StringType(), nullable=False)]
    + [StructField(col, FloatType(), nullable=False) for col in model_feature_names]
)

df_shap_values = (
    spark.createDataFrame(
        df_test.reindex(columns=model_feature_names + [cfg.student_id_col])
    )  # noqa: F821
    .repartition(sc.defaultParallelism)  # noqa: F821
    .mapInPandas(
        ft.partial(
            modeling.inference.calculate_shap_values_spark_udf,
            student_id_col=cfg.student_id_col,
            model_features=model_feature_names,
            explainer=explainer,
            mode=train_mode,
        ),
        schema=shap_schema,
    )
    .toPandas()
    .set_index(cfg.student_id_col)
    .reindex(df_test[cfg.student_id_col])
    .reset_index(drop=False)
)
```

### DIFF
--- a/src/student_success_tool/modeling/inference.py
+++ b/src/student_success_tool/modeling/inference.py
@@ -150,12 +150,12 @@ def calculate_shap_values(
         https://shap.readthedocs.io/en/stable/generated/shap.KernelExplainer.html
     """
     # preserve student ids
-    student_ids = df.loc[:, student_id_col]
+    student_ids = df.loc[:, student_id_col].reset_index(drop=True)
     # impute missing values and run shap values using just features
     features_imp = df.loc[:, feature_names].fillna(fillna_values)
-    shap_values = explainer.shap_values(features_imp)
+    explanation = explainer(features_imp)
     return (
-        pd.DataFrame(data=shap_values, columns=feature_names)
+        pd.DataFrame(data=explanation.values, columns=explanation.feature_names)
         # reattach student ids to their shap values
         .assign(**{student_id_col: student_ids})
     )
--- a/tests/modeling/test_inference.py
+++ b/tests/modeling/test_inference.py
@@ -1,3 +1,5 @@
+import collections
+
 import numpy as np
 import pandas as pd
 import pytest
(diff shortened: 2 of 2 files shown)


crysg (commented):
Nice! Seems fine to me, though @pemujo knows far more about this than I do, so will defer to him.

@pemujo, @samroon2 -- would this be where in the long term DK would be saving the shap explainers?

vishpillai123 (approved):
Looks like you're using the explainer directly similar way as the fellows instead of saving shap_values or feature_names as separate objects. Looks good!

pemujo:
Great, I will update the pipeline to use this function. Thanks @bdewilde @vishpillai123

src/student_success_tool/modeling/inference.py
```diff
@@ -150,12 +150,12 @@ def calculate_shap_values(
...
     # preserve student ids
-    student_ids = df.loc[:, student_id_col]
+    student_ids = df.loc[:, student_id_col].reset_index(drop=True)
     # impute missing values and run shap values using just features
     features_imp = df.loc[:, feature_names].fillna(fillna_values)
-    shap_values = explainer.shap_values(features_imp)
+    explanation = explainer(features_imp)
     return (
-        pd.DataFrame(data=shap_values, columns=feature_names)
+        pd.DataFrame(data=explanation.values, columns=explanation.feature_names)
```
pemujo, line 158: Hi @bdewilde @vishpillai123 I'm just checking this now, and this won't work for the shap.plots.beeswarm because beeswarm accepts Explanation object, not the np array with the values only (which is returned by explanation.values) nor the dataframe built from the np array.

bdewilde: I'm aware, that's what my question in the PR itself is about. I don't know how to combine multiple `Explanation` objects when running SHAP in a distributed fashion, or if it's even possible.

PR #100, opened 2025-03-09
extend utils refactoring

bdewilde:
## changes

incorporates pipeline utils into existing src code

## context

building on work started in PR #99 

this `pipeline_utils` subpackage was added to package haphazardly to address a technical issue in databricks; this is an attempt to more fully integrate the functionality

## questions

- this breaks the goog fellows' pipeline code! shall i update that in this PR, or will that require some extra work? i'm not super familiar with the setup there..
- the databricks runtime includes various google packages (see screenshot), but not `google-cloud`. 
image

on the `google-cloud` pypi page, i see this warning:

> WARNING: The google-cloud Python package is deprecated. On June 18, 2018, this package will no longer install any other packages. Please install the [product-specific google-cloud-* packages](https://github.com/GoogleCloudPlatform/google-cloud-python#google-cloud-python-client) needed for your application.

so, i've taken the liberty of updating the dependency to `google-cloud-storage`, and one that's aligned with the databricks runtime. does this work for y'all?


vishpillai123 (approved):
nice job combining google fellows work with ours!

bdewilde:
Thanks for the approval @vishpillai123 ! However, I don't want to break the google folks' pipeline code w/o their permission (or lining up the necessary fixes to be merged alongside these). @pemujo and @samroon2 , could you take a look here and advise?

pemujo:
Hi @bdewilde Thanks for checking. Yes, it is ok to remove the old google-cloud deprecated library. It would not break the pipeline since it is not really used anymore. We are only using [Python Client for Google Cloud Storage](https://pypi.org/project/google-cloud-storage/) in the pipeline.

PR #101, opened 2025-03-09
[pdp] improve term features

bdewilde:
## changes

- adds "core terms" input param for feature engineeering, and corresponding field in project config
- replaces "fall-spring" term features with configurable ^ "core" term features, which defaults to FALL/SPRING (suitable for semester-based calendars)
- adds additional "non-core" term feature, to account for students that enroll in e.g. WINTER/SUMMER terms

## context

Noreen asked about differentiating between core and non-core terms (she called them "major" terms, but I didn't want to overload schools' terminology wrt "major") in the context of targets; it reminded me that I wanted to do this in the feature engineering context.

follow-up to PR #88 

## questions

- is the core/non-core terminology okay? if not, is there a better term? i did some web recon and found no consistent naming for terms that are considered a core part of the academic year as opposed to those that are supplemental.


nm3224:
@bdewilde this looks good to me - do you think you can adjust the targets to account for this too (particularly when selecting the "nth" term? What I'm doing right now is just excluding summer terms the same way we exclude pre-cohort terms before calling the targets function, so it ignores them, but something tells me that's not ideal best practice...

nm3224 (approved):

bdewilde:
@nm3224 changes to targets is totally separate from this work, let's not blow up scope unnecessarily here

PR #102, opened 2025-03-10
Have gcsutils.py handle subdirs

crysg:
There is a top-level directory in the output directory which we don't want to copy over (we just want the .png file in the top level directory).

## changes
Only copy the file when it's a .png file in the top-level directory.


vishpillai123:
hey @crysg, @bdewilde actually put pipeline_utils into utils so there may be a merge conflict from that [PR](https://github.com/datakind/student-success-tool/pull/100) with this one. We haven't yet merged that branch. You may need to refactor what you have proposed here.

vishpillai123 (approved):
Approving based on meeting discussion

PR #103, opened 2025-03-11
Revert "[inference validation] ensuring top features & shap values are descending"

vishpillai123:
Reverts datakind/student-success-tool#96.

Just discussed this with Pedro from Google fellows. Unfortunately, because for the synthetic data we're essentially modeling noise, the shap values _will_ be invalid (all zeros etc).  

This validation logic is working as intended, but it's maybe too premature to add this in the context of the fellows' deliverable. 

TODO: Add this logic (and other important checks) in separate automated inference validation check.


bdewilde (approved):
Bummer!

PR #111, opened 2025-03-19
recoding pell to T/F from Y/N

nm3224:
## changes
Recodes `pell_status_first_year` from Y/N -> T/F

## context
Doing this so we can evaluate feature importance when interpreting SHAP value impact on model 

## questions
Do we want to just re-code the original column (`pell_status_first_year`) here or make a new one and call it something else, ex. pell_grant_recipient? 
Not sure if the way I did it is most efficient so open to edits - thanks!


src/student_success_tool/features/pdp/student.py
```diff
@@ -62,6 +62,11 @@ def add_features(
             )
             for yr in credits_years
         },
+        # not sure if we want to just re-code the original column here or make a new one and call it something else, ex. pell_grant_recipient
```
bdewilde, line 65: how about we call it `student_is_pell_recipient_first_year`, for consistency with the other student-level features? and to follow the existing pattern, let's rename the `change_pell_to_T_F` function to the same thing.

src/student_success_tool/features/pdp/student.py
```diff
@@ -82,3 +87,12 @@ def diff_gpa_term_1_to_year_1(
...
 ) -> pd.Series:
     return df[year_col].sub(df[term_col])
+
+
+def change_pell_to_T_F(
+    df: pd.DataFrame, 
+    *, 
+    pell_col: str = "pell_status_first_year",
+) -> pd.Series:
+    # Convert 'Y' to True and 'N' to False
```
bdewilde, line 97: nit: no need for this code comment :)

src/student_success_tool/features/pdp/student.py
```diff
@@ -82,3 +87,12 @@ def diff_gpa_term_1_to_year_1(
...
+    return df[pell_col].map({'Y': True, 'N': False})
```
bdewilde, line 98: This is a fine way to do it (`Series.replace()` is the other). Since we know that inputs are required to be one of Y/N/null, this is safe. That said, let's lint this line and specify the dtype, to avoid surprises down the line:

```suggestion
    return df[pell_col].map({"Y": True, "N": False}).astype("boolean")
```

bdewilde (changes requested):
Left a couple comments with easy changes. The only notable thing you're missing is unit tests -- specifically, you'll probably need to add this field into existing unit tests.

nm3224:
> Left a couple comments with easy changes. The only notable thing you're missing is unit tests -- specifically, you'll probably need to add this field into existing unit tests.

@bdewilde got it, will work on that now! Already incorporated the simple stuff

PR #112, opened 2025-03-19
adding to EDA notebook

nm3224:
Updating EDA notebook 

## context & changes 
- added code to see how many students filtered out due to missing course identifiers are transfers vs. non-transfers
- added code to see how many students remain after we filter the above ^ out
- filtering cohort data as well for consistency with course_filtered dataset
- plotting first year GPA by enrollment type and credential type sought to evaluate if these types should be included in same or different models based on differences
- adding code for computing and plotting average and median values for percent credits earned over years (1-4)
- Adding a student-demographics section with some more detailed plots

## questions
Anything you guys would like to change?

### DIFF
--- a/pipelines/institution_id/00-data-assessment-TEMPLATE.py
+++ b/pipelines/institution_id/00-data-assessment-TEMPLATE.py
@@ -401,9 +401,11 @@ df_course.loc[
 # how many for transfers vs non transfers?
 missing_course = df_course.groupby("student_guid").filter(
     lambda x: x["course_prefix"].isna().all()
-)
+    )
 
-missing_course.groupby('student_guid')['enrolled_at_other_institution_s'].first().value_counts(dropna=False)
+missing_course.groupby('student_guid')[
+    'enrolled_at_other_institution_s'
+    ].first().value_counts(dropna=False)
 
 # COMMAND ----------
 
--- a/pipelines/institution_id/00-data-assessment-TEMPLATE.py
+++ b/pipelines/institution_id/00-data-assessment-TEMPLATE.py
@@ -382,6 +382,7 @@ _ = msno.matrix(
 
 # COMMAND ----------
 
+# how many total?
 num_missing_course_id = df_course["course_prefix"].isna().sum()
 pct_missing_course_id = 100 * num_missing_course_id / len(df_course)
 print(
@@ -394,6 +395,12 @@ print(
 df_course.loc[
     df_course["course_prefix"].isna(), "enrolled_at_other_institution_s"
 ].value_counts()
+
+# COMMAND ----------
+
+# how many for transfers vs non transfers?
+missing_course = df_course.groupby("student_guid").filter(lambda x: x["course_prefix"].isna().all())
+print(missing_course.groupby('student_guid')['enrolled_at_other_institution_s'].first().value_counts(dropna=False))
 
 # COMMAND ----------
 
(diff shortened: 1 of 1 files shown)


pipelines/institution_id/00-data-assessment-TEMPLATE.py
```diff
@@ -397,6 +398,16 @@
 
 # COMMAND ----------
 
+# how many for transfers vs non transfers?
+missing_course = df_course.groupby("student_guid").filter(
+    lambda x: x["course_prefix"].isna()
+    .all()
+    )
```
bdewilde, line 405: nit: style

```suggestion
    lambda x: x["course_prefix"].isna().all()
)
```

pipelines/institution_id/00-data-assessment-TEMPLATE.py
```diff
@@ -539,6 +550,17 @@
 
 # COMMAND ----------
 
+# how many students remain?
+df_course_valid['student_guid'].nunique()
```
bdewilde, line 554: did you mean `df_course_filtered` ?

pipelines/institution_id/00-data-assessment-TEMPLATE.py
```diff
@@ -590,6 +612,102 @@
...
+)
+
+# Set the ylabel
+ax.set(ylabel="Avg. GPA (Year 1)")
+
+# Move the legend to a different location (e.g., upper left)
+ax.legend(loc="lower left", title="Enrollment Type")
+
+# Show the plot
+plt.show()
```
bdewilde, line 635: no need for code comments that simply describe the code being executed :)

```suggestion
ax.set(ylabel="Avg. GPA (Year 1)")
ax.legend(loc="lower left", title="Enrollment Type")
```

pipelines/institution_id/00-data-assessment-TEMPLATE.py
```diff
@@ -590,6 +612,102 @@
...
+)
+
+# Set the ylabel
+ax.set(ylabel="Avg. GPA (Year 1)")
+
+# Move the legend to a different location (e.g., upper left)
+ax.legend(loc="lower left", title="Enrollment Intensity")
+
+# Show the plot
+plt.show()
```
bdewilde, line 659:
```suggestion
ax.set(ylabel="Avg. GPA (Year 1)")
ax.legend(loc="lower left", title="Enrollment Intensity")
```

pipelines/institution_id/00-data-assessment-TEMPLATE.py
```diff
@@ -643,6 +768,37 @@
 
 # COMMAND ----------
 
+# degree seeking by enrollment type
+
+(
+    sb.histplot(
```
bdewilde, line 774: I'm not sure `histplot` is the right plot -- possibly `countplot()` or `barplot()` instead? It looks like this is what we're using elsewhere, though, so presumably I had good reasons for it back when I implemented...

bdewilde (commented):
did a quick pass -- looks mostly fine! though i think you have to tidy the code to make the linter happy...

pipelines/institution_id/00-data-assessment-TEMPLATE.py
nm3224: Yes, good catch! Thanks

nm3224: I think it was formerly called "valid"

pipelines/institution_id/00-data-assessment-TEMPLATE.py
nm3224: I agree, I just defaulted to histplot to follow precedent (which you may have had good reason for)!

nm3224:
> did a quick pass -- looks mostly fine! though i think you have to tidy the code to make the linter happy...

@bdewilde have we figured out best practice for linting? I used to run things in terminal and it would work but I'm getting a weird bug when doing that now...

nm3224:
> > did a quick pass -- looks mostly fine! though i think you have to tidy the code to make the linter happy...
> 
> @bdewilde have we figured out best practice for linting? I used to run things in terminal and it would work but I'm getting a weird bug when doing that now...

@bdewilde okay linter is working now - that was so painful, I didn't realize I had random white spaces and how specific it is about double "" vs single '' quotes! All done now :)

bdewilde (approved):

PR #117, opened 2025-03-24
docs: update readme setup instructions

bdewilde:
## changes

updates docs to current practice

## context

https://datakind.slack.com/archives/C07RJHJU59D/p1742844952640059?thread_ts=1742053501.136229&cid=C07RJHJU59D


pemujo:
I would not include runtime 16.X until the 16.X LTS version is released. It should be coming out pretty soon since Databricks release new version every 6 months aproximately and 15.4 LTS came in August 2024. 
I also would not use 16.X until you get clarification from Databricks about why they have an unsupported mix of Python 3.12.x and Pandas 1.5.3

bdewilde:
> I would not include runtime 16.X until the 16.X LTS version is released. It should be coming out pretty soon since Databricks release new version every 6 months aproximately and 15.4 LTS came in August 2024. I also would not use 16.X until you get clarification from Databricks about why they have an unsupported mix of Python 3.12.x and Pandas 1.5.3

I'm describing what is currently done -- changing current practice is a different conversation, I think. :)

Also, pandas v1.5.3 absolutely does work in Python 3.12, it just doesn't have wheels pre-built for it because it was released before PY3.12. In the package's metadata, they only require a minimum version of Python, and don't set a maximum.

kaylawilding (approved):

README.md
```diff
@@ -47,9 +47,9 @@ Please read the [CONTRIBUTING](CONTRIBUTING.md) to learn how to contribute to th
 
 ### databricks notebook
 
-1. Connect notebook to a cluster running Databricks Runtime [14.3 LTS](https://docs.databricks.com/en/release-notes/runtime/14.3lts.html) or [15.4 LTS](https://docs.databricks.com/en/release-notes/runtime/15.4lts.html).
+1. Connect notebook to a cluster running Databricks Runtime [15.4 LTS](https://docs.databricks.com/en/release-notes/runtime/15.4lts.html) or [16.x](https://docs.databricks.com/aws/en/release-notes/runtime/16.2).
```
kaylawilding, line 50: Should we specify 15.4 for training and 16.2 can be used for other steps? Also, maybe we should just hold off on using 16.2 until 16.3 LTS is released? We can also just update this later as we make decisions and this is fine for now.

bdewilde: I think we need to hold on that -- for recent context, see this thread (on which you've been tagged) https://datakind.slack.com/archives/C07DWDMN8SW/p1742606642838799

bdewilde: Going to merge, but all good to revisit once this Databricks env stuff gets wrangled :/

PR #118, opened 2025-03-25
adding FNPR bias evaluation and flagging system

vishpillai123:
We've revamped our bias evaluation to include the metric False Negative Parity Rate (FNPR) along with a flagging system that is currently being evaluated across our school portfolio.

## changes
- Changed train template to include bias evaluation along with an iteration across top X runs of experiment post-training. X can be changed depending on how many runs you want to evaluate. 
- Added `bias_detection.py` file under modeling source folder. I chose the word "detection" here since we already have an `evaluation.py` file. 
- Added unit tests for `bias_detection` file.
- The new bias evaluation involves: 
1. Calculate FNPR within each subgroup with a minimum sample threshold for FN or TP
2. Calculate Confidence Intervals & p-values for FNPR.
3. Compare subgroup to subgroup within each group in order to detect potential bias.

## context
- Our model transparency efforts are underway and start with formalized and rigorous bias evaluation.
- This process will be very iterative as we learn more from schools.
- As opposed to just evaluating the best run, I've added iteration across top X runs because often the best run may have more biased flags than other runs. I've defaulted to 3 here, but in my experience, looking up to 10 runs can be useful if your dataset is challenging (i.e. imbalanced, etc) and your models are prone to bias.

## questions
- I've reviewed this in our DS team meetings, so let me know if you have any issues with the implementation or where I put code. 
- Lots of TODOs  below for future work... if you have suggestions on more TODOs, please share. 
  
  - TODO: Add confidence intervals for difference in outcome prevalence. Often, we are interested in FNPR (or recall if we don't have enough samples) with more information on outcome prevalence as it can inform whether our model is under-predicting or over-predicting a particular subgroup. This way we have perspective on systemic bias (outcome prevalence) and FNPR (decision fairness).
  - TODO: Find alternatives to FNPR for low sample sizes of FN < 50 and TP < 50.
  - TODO: Add visualization for model card work.
  - TODO: Add automated bucketing of subgroups. Example: Bucket Age 40-49, 50-59, 60+ as 40+. This may be challenging to implement but could also be necessary for our model cards effort when we do not have enough samples for certain subgroups.


vishpillai123:
Tagging @Mesh-ach as primary reviewer (as per alignment with Rachel), but tagging others to review if they would like to! Especially since this is a pretty big change to our normal process.

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+MODERATE_FLAG_THRESHOLD = 0.1
+LOW_FLAG_THRESHOLD = 0.05
+
+# Define flag types
+FLAG_NAMES = {
+    "🟢 NO BIAS": "no_bias",
+    "🟡 LOW BIAS": "low_bias",
+    "🟠 MODERATE BIAS": "moderate_bias",
+    "🔴 HIGH BIAS": "high_bias",
+}
```
bdewilde, line 21: fancy! ✨

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+    min_fnpr_samples: int = MIN_FNPR_SAMPLES,
+) -> tuple[float, float, float]:
+    """
+    Calculates the False Negative Prediction Rate (FNPR) and its confidence interval.
+
+    Args:
+        targets: Labels from model output
+        preds: Predictions from model output
+        min_fnpr_samples: Minimum number of true positives or false negatives for FNPR calculation. When TP or FN are low, FNPR can be very unstable and flag subgroups based on small differences in TP or FN.
+    Returns:
```
bdewilde, line 36: nit: put an empty line between "args" and "returns" block

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+    "🟢 NO BIAS": "no_bias",
+    "🟡 LOW BIAS": "low_bias",
+    "🟠 MODERATE BIAS": "moderate_bias",
+    "🔴 HIGH BIAS": "high_bias",
+}
+
+
+def calculate_fnpr_and_ci(
+    targets: pd.Series,
+    preds: pd.Series,
```
bdewilde, line 26: This is potentially confusing. I'm assuming `targets` represents the "true" / known outcomes, while `preds` is the model predictions? Are those going to be binary True/False or probabilities? Describing in the docstring would be a big help.

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+        min_fnpr_samples: Minimum number of true positives or false negatives for FNPR calculation. When TP or FN are low, FNPR can be very unstable and flag subgroups based on small differences in TP or FN.
+    Returns:
+        Tuple of FNPR, lower CI bound, and upper CI bound.
+    """
+    cm = confusion_matrix(targets, preds, labels=[False, True])
+    tn, fp, fn, tp = (
+        cm.ravel()
+        if cm.shape == (2, 2)
+        else np.bincount(np.array(targets) * 2 + np.array(preds), minlength=4)
+    )
```
bdewilde, line 44: this is tricksy :) do we have a unit test to handle the case when `cm.shape != (2, 2)`? and in which situations is that the case?

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+}
+
+
+def calculate_fnpr_and_ci(
+    targets: pd.Series,
+    preds: pd.Series,
+    min_fnpr_samples: int = MIN_FNPR_SAMPLES,
+) -> tuple[float, float, float]:
+    """
+    Calculates the False Negative Prediction Rate (FNPR) and its confidence interval.
```
bdewilde, line 30: would be nice to have a link out to more info about this metric in a "references" block or something 🙏

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+    )
+
+    denominator = fn + tp
+
+    if (tp < min_fnpr_samples) or (fn < min_fnpr_samples):
+        return np.nan, np.nan, np.nan
+
+    fnpr = fn / denominator
+    margin = Z * np.sqrt((fnpr * (1 - fnpr)) / denominator)
+    return fnpr, max(0, fnpr - margin), min(1, fnpr + margin)
```
bdewilde, line 53: full disclosure: I'm assuming all this math is correct

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+) -> bool:
+    """
+    Checks whether confidence intervals (CIs) overlap. If they do, the FNPR differences
+    are within the margin of error at the 95% confidence level. If the CIs do not
+    overlap, this suggests strong statistical evidence that the FNPRs are different.
+
+    Args:
+        ci1: Confidence interval (min, max) for subgroup 1
+        ci2: Confidence interval (min, max) for subgroup 2
+    Returns:
```
bdewilde, line 68: nite: add whitespace line plz

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+        Boolean indicating whether the CIs overlap.
+    """
+    return not (ci1[1] < ci2[0] or ci2[1] < ci1[0])
+
+
+def z_test_fnpr_difference(
+    fnpr1: float,
+    fnpr2: float,
+    denominator1: int,
+    denominator2: int,
```
bdewilde, line 78: "denominator" is vague, I think this would be clearer with variables named like `num_negatives` or something

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    if (
+        denominator1 <= 30 or denominator2 <= 30
+    ):  # Ensures valid sample sizes for z-test
+        return np.nan
+    std_error = np.sqrt(
+        (fnpr1 * (1 - fnpr1)) / denominator1 + (fnpr2 * (1 - fnpr2)) / denominator2
+    )
+    z_stat = (fnpr1 - fnpr2) / std_error
+    return float(2 * (1 - st.norm.cdf(abs(z_stat))))  # Two-tailed p-value
```
bdewilde, line 102: I _think_ this calculation is preferred:

```suggestion
    return 2.0 * st.norm.sf(abs(z_stat))  # Two-tailed p-value
```

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+    Returns:
+        Two-tailed p-value for the z-test for the FNPR difference between the two subgroups.
+    """
+    if (
+        denominator1 <= 30 or denominator2 <= 30
+    ):  # Ensures valid sample sizes for z-test
+        return np.nan
+    std_error = np.sqrt(
+        (fnpr1 * (1 - fnpr1)) / denominator1 + (fnpr2 * (1 - fnpr2)) / denominator2
+    )
```
bdewilde, line 100: i understand that this could be worked out by following Python's order of operations when multiple mathematical operators are acting together in an expression, but... could you just add some extra parens to group calculations together more obviously?

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+    z_stat = (fnpr1 - fnpr2) / std_error
+    return float(2 * (1 - st.norm.cdf(abs(z_stat))))  # Two-tailed p-value
+
+
+def log_bias_flag(
```
bdewilde, line 105: confused by this function name

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+        dataset: Name of the dataset (e.g. train/test/validate)
+        flag: Flag value (e.g. "🔴 HIGH BIAS", "🟠 MODERATE BIAS", "🟡 LOW BIAS", "🟢 NO BIAS")
+        p_value: p-value for the z-test for the FNPR difference of the subgroup pair
+    Returns:
+        Dictionary containing bias flag information.
+    """
+    flag_entry = {
+        "group": group,
+        "subgroups": f"{subgroup1} vs {subgroup2}",
+        "difference": fnpr_diff * 100,
```
bdewilde, line 133: why multiply the fnpr diff by 100 and then name it "difference"?

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+        "type": bias_type
+        if np.isnan(p_value)
+        else f"{bias_type}, p-value: {'< 0.001' if p_value < 0.001 else f'{p_value:.3f}'}",
```
bdewilde, line 136: hard to read

```suggestion
        "type": (
            bias_type
            if np.isnan(p_value)
            else f"{bias_type}, p-value: {'< 0.001' if p_value < 0.001 else f'{p_value:.3f}'}"
        ),
```

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    Aggregate bias flag information for a given subgroup pair into a dict.
+
+    Args:
+        group: Name of the group (e.g. "Gender", "Race", "Age")
+        subgroup1: Name of the subgroup 1 (e.g. "Female", "Male", "Asian", "African American", "Caucasian")
+        subgroup2: Name of the subgroup 2 (e.g. "Female", "Male", "Asian", "African American", "Caucasian")
+        fnpr_diff: Absolute value of difference in FNPR
+        bias_type: Type of bias (e.g. "Non-overlapping CIs", "Overlapping: : p-value: ...")
+        dataset: Name of the dataset (e.g. train/test/validate)
```
bdewilde, line 124: this could be confusing, since datasets themselves are named, but here you seem to be talking about "splits"

bdewilde (commented):
added some comments, but deferring a more thorough review to @Mesh-ach

pipelines/institution_id/02-train-model-TEMPLATE.py
```diff
@@ -227,114 +222,179 @@
...
+    if optimization_metric in ["log_loss", "roc_auc"]
+    else f"metrics.val_{optimization_metric}_score"
+)
+
+# Sort and select top run IDs based on min/max with loss vs. score
+ascending_order = optimization_metric == "log_loss"
+top_run_ids = (
+    runs.sort_values(by=search_metric, ascending=ascending_order)
+    .iloc[:top_x]["run_id"]
+    .tolist()
```
bdewilde, line 247: actually, i have a general question about the changes in the training nb: why are you running this over the top N scoring models, rather than just the best? it's hard to see what's changed in the git diff, so just... broadly unsure about what's happening and why here.

bdewilde: ugh, nm, you mostly addressed in your PR message. the current eval code/loop feels like it needs to be cleaned up, but that can certainly be down outside of this PR.

src/student_success_tool/modeling/bias_detection.py
Mesh-ach: ✚

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
+import pandas as pd
+import numpy as np
+import scipy.stats as st
+from sklearn.metrics import confusion_matrix
+
+# FNPR sample threshold
+MIN_FNPR_SAMPLES = 50
```
Mesh-ach, line 7: Should the Min FNPR samples be adaptable to each institution dataset, or are they like a fixed stable value determined as the best minimum through testing?

src/student_success_tool/modeling/bias_detection.py
Mesh-ach: Since most of our work is on binary classification anyway, wouldn't cm always be 2x2 for our use case?

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -0,0 +1,213 @@
...
+
+    Args:
+        fnpr1: FNPR value for subgroup 1
+        fnpr2: FNPR value for subgroup 2
+        denominator1: Number of false negatives + true negatives for subgroup 1
+        denominator2: Number of false negatives + true negatives for subgroup 2
+    Returns:
+        Two-tailed p-value for the z-test for the FNPR difference between the two subgroups.
+    """
+    if (
```
Mesh-ach, line 94: do you think we might wanna add an assertion here instead of returning nans, especially since this might be integrated into our standard pipeline and wouldn't invalids returned?

vishpillai123: Invalids won't be returned. FNPR will just not be calculated. This is the case for small sample sizes for subgroups.

src/student_success_tool/modeling/bias_detection.py
vishpillai123: This is a fixed stable value that I have found that works with the thresholds that we've all decided on 5%, 10%, 15% FNPR difference etc. 

The issue more comes from that FNPR is very unstable below 30 samples, and above 50 is where more stable and we can utilize these decision boundaries with confidence. Otherwise, we won't flag for FNPR.

src/student_success_tool/modeling/bias_detection.py
vishpillai123: It's just a percentage difference. But sure, I can call it percentage difference.

src/student_success_tool/modeling/bias_detection.py
vishpillai123: We are logging the bias flags as an ML artifact.

src/student_success_tool/modeling/bias_detection.py
vishpillai123: I was running into this issue with my first school I was testing, but now it seems to work fine without this logic, so I will take it out.

src/student_success_tool/modeling/bias_detection.py
vishpillai123: We could potentially create more adaptable thresholding here. My intention behind this is to flag statistical abnormalities in bias. If a school has a small dataset, it will basically flag everywhere unless we create some type of thresholding.

Perhaps, in the future, we can have adaptable thresholds for high bias, low bias, etc. that is related to the number of samples. This is definitely future work at the moment.

vishpillai123:
changes are implemented, let me know what you think!

Mesh-ach (approved):

src/student_success_tool/modeling/bias_detection.py
bdewilde: I stand by my comment -- that reads to me as "potentially confusing" :)

src/student_success_tool/modeling/bias_detection.py
bdewilde: got it, thanks! personally i don't care for mlflow's (mis)use of the term "log" ...

PR #123, opened 2025-03-30
[pdp] update project config template

bdewilde:
## changes

- follows naming convention for datasets and models
- adds more commonly used fields, even if commented out
- fills in placeholder values with slightly more realistic values

## context

there's been some confusion about how to translate this template into an actual project config; these changes are meant to make things more clear and closer to "standard"

## questions

- anything you don't like here?
- anything you would like included here?

### DIFF
--- a/pipelines/institution_id/config-TEMPLATE.toml
+++ b/pipelines/institution_id/config-TEMPLATE.toml
@@ -1,7 +1,7 @@
 institution_id = "INST_ID"
 institution_name = "INST NAME"
 
-student_id_col = "student_guid"
+student_id_col = "student_id"
 target_col = "target"
 split_col = "split"
 sample_weight_col = "sample_weight"
@@ -11,10 +11,17 @@ pred_prob_col = "pred_prob"
 pos_label = true
 random_state = 12345
 
-[datasets.labeled]
+[datasets.ingested_YYYYMMDD]
 raw_course = { file_path = "/Volumes/CATALOG/INST_NAME_bronze/INST_NAME_bronze_file_volume/FILE_NAME_COURSE.csv" }
 raw_cohort = { file_path = "/Volumes/CATALOG/INST_NAME_bronze/INST_NAME_bronze_file_volume/FILE_NAME_COHORT.csv" }
-preprocessed = { table_path = "CATALOG.SCHEMA.TABLE_NAME" }
+preprocessed = { table_path = "CATALOG.INST_NAME_silver.DATASET_NAME_preprocessed" }
+# TODO: once PR #122 is merged
+# finalized = { table_path = "CATALOG.INST_NAME_gold.INST_ID_advisor_output_ETC_YYYYMMDD" }
+
+[models.TARGET_CHECKPOINT]
+experiment_id = "EXPERIMENT_ID"
+run_id = "RUN_ID"
+framework = "sklearn"
 
 [preprocessing]
 splits = { train = 0.6, test = 0.2, validate = 0.2 }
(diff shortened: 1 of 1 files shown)


pipelines/institution_id/config-TEMPLATE.toml
```diff
@@ -11,10 +11,17 @@ pred_prob_col = "pred_prob"
 pos_label = true
 random_state = 12345
 
-[datasets.labeled]
+[datasets.ingested_YYYYMMDD]
```
vishpillai123, line 14: Is this meant to cover when schools send us new data for inference or when they send us updated data for training?

EIther way, naming the datasets "ingested_YYYYMMDD" would require us to change the notebooks every time schools send us new data, right? I know based on our discussion in team weekly last week, we want to record dates of files sent by the institution, so I think I understand the intention here. Though, I think a better practice to make this more programmable is leaving how we had it before "labeled" and "unlabeled". Otherwise, our configs are going to get messy over time, and this is also going to be far too manual (in my opinion) to update configs.

pipelines/institution_id/config-TEMPLATE.toml
```diff
@@ -11,10 +11,17 @@ pred_prob_col = "pred_prob"
...
 raw_course = { file_path = "/Volumes/CATALOG/INST_NAME_bronze/INST_NAME_bronze_file_volume/FILE_NAME_COURSE.csv" }
 raw_cohort = { file_path = "/Volumes/CATALOG/INST_NAME_bronze/INST_NAME_bronze_file_volume/FILE_NAME_COHORT.csv" }
-preprocessed = { table_path = "CATALOG.SCHEMA.TABLE_NAME" }
+preprocessed = { table_path = "CATALOG.INST_NAME_silver.DATASET_NAME_preprocessed" }
+# TODO: once PR #122 is merged
+# finalized = { table_path = "CATALOG.INST_NAME_gold.INST_ID_advisor_output_ETC_YYYYMMDD" }
+
+[models.TARGET_CHECKPOINT]
```
vishpillai123, line 21: I think it's more intuitive for this to be below preprocessing and modeling.training (how we had it before), but yeah not a big deal either way.

pipelines/institution_id/config-TEMPLATE.toml
bdewilde: It works for _both_ training and inference data, since both are datasets with the same configuration options. imo this pattern doesn't lead to messy config files! It also doesn't require that the notebooks be updated; you can specify `dataset_name` as a job config.

pipelines/institution_id/config-TEMPLATE.toml
bdewilde: I changed the ordering so that "artifacts" are consolidated up toward the top, while "process" configuration is lower down. Agree that it's not a big deal either way, this just seemed easier in practice.

pipelines/institution_id/config-TEMPLATE.toml
nm3224: @bdewilde did you remove the "predictions" artifact representing the modeling dataset with the labels or is that just not here b/c it was unedited/left alone?

bdewilde: hey @nm3224 , "predictions" was originally intended for just the raw model predictions, which (afaik) we've not usually saved into unity catalog. in this PR i added the previously-tbd "finalized" step of each dataset, which is intended for external-facing results sent back to clients

vishpillai123 (approved):

PR #126, opened 2025-04-03
Add bias visualization

vishpillai123:

### DIFF
--- a/pipelines/institution_id/02-train-model-TEMPLATE.py
+++ b/pipelines/institution_id/02-train-model-TEMPLATE.py
@@ -225,8 +225,6 @@ else:
 
 # COMMAND ----------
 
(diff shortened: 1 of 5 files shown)


bdewilde:
I see this is still in draft, but whenever it's ready for review, could you include an example plot in the PR message? (You can just drag-and-drop a file from disk onto the input field, it should automatically convert into valid markdown.)

bdewilde:
closed??

vishpillai123:
> closed??

lol I was having trouble rebasing and databricks kept showing diffs, so I just chose to do a hard reset and recreate the branch. I've had to do this a couple times with databricks cuz I'll reset the branch but then the file I'm on keeps the previous changes. It's weird.

PR #127, opened 2025-04-04
[pdp] improve model evaluation code

bdewilde:
## changes

- adds "evaluation" block to project config, and puts a single parameter in it; not very useful rn, but could be in the future, and at the very least gets rid of a "magic" number in the training model nb
- improves dataset block in template project config; this is kind of unrelated, but it snuck in (sorry)
- adds function to compute many classification model evaluation metrics; should be used to simplify the eval loop in model training nb
- splits feature permutation importance into two functions: one to compute the importances, and the other to plot them; this gives people more control
- renames / reorders various evaluation plotting functions, for consistency

**out-of-scope:** i didn't update the training model nb, because Vish is currently in that space still making changes for bias evaluation. we should do all those together, as a fast-follow, and the last bit before releasing v0.2

## context

our model evaluation code is messy and inconsistent; this takes a couple steps in a good direction

### DIFF
--- a/src/student_success_tool/modeling/evaluation.py
+++ b/src/student_success_tool/modeling/evaluation.py
@@ -310,6 +310,47 @@ def compare_trained_models(
 
 
 def compare_trained_models_plot(
+def compute_feature_permutation_importance(
+    model: sklearn.base.BaseEstimator,
+    features: pd.DataFrame,
+    target: pd.Series,
+    *,
+    n_repeats: int = 5,
+    scoring: t.Optional[str | t.Callable] = None,
+    sample_weight: t.Optional[np.ndarray] = None,
+    random_state: t.Optional[int] = None,
+) -> sklearn.utils.Bunch:
+    """
+    Compute model features' permutation importance.
+
+    Args:
+        model
+        features
+        target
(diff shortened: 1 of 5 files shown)


src/student_success_tool/modeling/evaluation.py
```diff
@@ -386,122 +391,187 @@ def compare_trained_models_plot(
...
+    ax.set_ylim(bottom=-0.05, top=1.05)
+    ax.legend(loc="lower right")
+    return fig
 
 
-def create_evaluation_plots_by_subgroup(
-    data: pd.DataFrame,
-    risk_score_col: str,
-    y_true_col: str,
+def plot_sla_curve(
```
vishpillai123, line 468: How useful do you find sla curves? I'm wondering if we really need them.

vishpillai123 (approved):

src/student_success_tool/modeling/evaluation.py
bdewilde: tbh I don't use them, but don't want to rock the boat too hard with this update :) We can revisit model eval more holistically after this next release?

PR #128, opened 2025-04-04
recreating PR because rebase was not working as it should with databr…

vishpillai123:
…icks

### DIFF
--- a/pipelines/institution_id/02-train-model-TEMPLATE.py
+++ b/pipelines/institution_id/02-train-model-TEMPLATE.py
@@ -253,7 +253,7 @@ top_run_ids = (
 for run_id in top_run_ids:
     with mlflow.start_run(run_id=run_id) as run:
         model = mlflow.sklearn.load_model(f"runs:/{run_id}/model")
-        df_all_flags = pd.DataFrame()
+        df_group_flags = pd.DataFrame()
         df_pred = df.assign(
             **{
                 cfg.pred_col: model.predict(df_features),
@@ -309,6 +309,7 @@ for run_id in top_run_ids:
                             {
                                 "group": group,
                                 "subgroup": subgroup,
+                                "split_name": split_name,
                                 "fnpr": fnpr,
                                 "ci": (fnpr_lower, fnpr_upper),
                                 "size": len(subgroup_data),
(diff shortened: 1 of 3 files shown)


pipelines/institution_id/02-train-model-TEMPLATE.py
```diff
@@ -373,38 +374,51 @@
...
+                    bias_subgroup_flags = [
+                        flag
+                        for flag in all_subgroup_flags
+                        if flag["flag"] not in ["🟢 NO BIAS", "⚪ INSUFFICIENT DATA"]
+                    ]
+                    for bias_flag in bias_subgroup_flags:
+                        logging.info(
+                            f"""Run {run_id}: {bias_flag["group"]} on {bias_flag["split_name"]} - {bias_flag["subgroups"]}, 
+                            FNPR Difference: {bias_flag["fnpr_percentage_difference"]}% ({bias_flag["type"]}) [{bias_flag["flag"]}]"""
+                        )
```
bdewilde, line 388: this is bad form wrt logging messages -- it's strongly recommended to use "%s" instead of f-strings, one of the very few places where the old-style string interpolation is still used

src/student_success_tool/modeling/evaluation.py
```diff
@@ -644,6 +644,59 @@ def plot_shap_beeswarm(shap_values):
     return plt.gcf()
 
 
+def fnpr_group_plot(fnpr_data: list) -> matplotlib.figure.Figure:
```
bdewilde, line 647: for consistency with my recent renaming:

```suggestion
def plot_group_fnpr(fnpr_data: list) -> matplotlib.figure.Figure:
```

src/student_success_tool/modeling/evaluation.py
```diff
@@ -644,6 +644,59 @@ def plot_shap_beeswarm(shap_values):
...
+    """
+    Plots False Negative Prediction Rate (FNPR) for a group by subgroup on
+    a split (train/test/val) of data with confidence intervals.
```
bdewilde, line 650: I noticed this earlier, and am wondering again now: Why do we do FNPR on each split separately? Seems like we could just use the full (un-split) labeled dataset.

src/student_success_tool/modeling/evaluation.py
```diff
@@ -644,6 +644,59 @@ def plot_shap_beeswarm(shap_values):
...
+
+    Parameters:
+    - fnpr_data: List with dictionaries for each subgroup. Each dictionary
+    comprises of group, fnpr, ci (confidence interval), and split_name keys.
```
bdewilde, line 654: google-style docstring, please!

```suggestion
    Args:
        fnpr_data: List with dictionaries for each subgroup. Each dictionary
            comprises of group, fnpr, ci (confidence interval), and split_name keys.
```

pipelines/institution_id/02-train-model-TEMPLATE.py
```diff
@@ -373,38 +374,51 @@
                         local_path=metrics_tmp_path, artifact_path="subgroup_metrics"
                     )
 
-                    bias_flags = modeling.bias_detection.flag_bias(
-                        fnpr_data, split_name
-                    )
+                    all_subgroup_flags = modeling.bias_detection.flag_bias(fnpr_data)
```
bdewilde, line 377: ngl the loops-within-loops-within-loops of all this is very hard to follow. we should definitely simplify this soon, partially leveraging the changes/additions I made in #127

bdewilde (approved):

src/student_success_tool/modeling/evaluation.py
vishpillai123: Yeah we talked about this on Thursday in the knowledge sharing meeting. I've personally had experiences running this code on my schools where an FNPR difference is much more pronounced on test vs. train or val. I assume it's due to assumptions that the model is making about feature distributions on training that is exacerbated on test. We definitely want to have visibility into this.

It also makes sense to run only on all splits since we are using this for model selection and would like to know the bias evaluation on train, test, and val (but obviously test is what matters the most and what we will show in the model cards).

vishpillai123:
@bdewilde just a heads up, since I'm going to have the refactoring PR here in the next few days or so, I'm going to apply your feedback to the next PR and get this one merged.

PR #129, opened 2025-04-04
refactor evaluation code for v2 release

vishpillai123:
## changes
- Cleaned up the 150+ lines block of code in training notebook and is now only ~30 lines of code.
- The previous large block of code is now captured by three functions: `load_model_and_predict`, `evaluate_performance`, and `evaluate_bias`. The first two are in evaluation.py, the third is in bias_detection.py.
- Applied feedback from PR #128: switched fnpr group plot -> plot fnpr group, removed the for loop inside for loop mess. There's still the same looping going on, but it's just cleaner with functions calling other functions. The main code (30ish lines) just loops over top runs. Also applied the other two comments from @bdewilde around google docstring format and cleaning up the logging statement to use %s instead of f-string.
- Removed the printing of plots in the console using `plt.close()` when calling `mlflow.log_figure()` since it gets really, really messy when you execute multiple runs (literally like 20+ figures are printed). Also realized this is not like absolutely necessary when you can just look at the MLflow run and look up the artifacts.
- Added functions to log metrics & plots to mlflow.
- Tested it on one of my schools to make sure everything works.
- Added unit tests where I thought it was needed, but if you would like more tests - please let me know.

## context
- We are releasing a new version of SST package and evaluation code needs a lot of cleanup.
- Separated out performance evaluation from bias evaluation. This is more to just keep things clean & working for now. There is some redundancy between the two files though, mostly with: `compute_classification_eval_metrics` in evaluation.py does a lot of same stuff as `compute_subgroup_bias_metrics` in bias_detection.py. Though, we also add FNPR and other more bias related code. This isn't super problematic, but just pointing out that this is some redundant code that we may want to clean up later. 

## questions
- Lots of cleanup and moving stuff around for both performance & bias evaluation, so any suggestions/feedback is welcome.
- Not a question, but just a heads up that evaluation can take a few mins. For example, evaluating performance & bias across top 5 runs takes around 7 minutes. Evaluating just the top run will take about 60-90 seconds. We can probably optimize this performance in the future.


src/student_success_tool/modeling/evaluation.py
```diff
@@ -64,6 +65,144 @@ def extract_training_data_from_model(
...
+        df_features (pd.DataFrame): DataFrame but with only columns from model training features.
+        pred_col (str): Column name for the predictions.
+        pred_prob_col (str): Column name for predicted probabilities.
+
+    Returns:
+        model (sklearn.base.BaseEstimator): The loaded model from experiment_id/run_id.
+        df_pred (pd.DataFrame): A copy of the original DataFrame `df`, with two new columns: pred_col: predicted
+        class labels and pred_probs_col: predicted probabilities for the positive class.
+    """
+    model = mlflow.sklearn.load_model(f"runs:/{run_id}/model")
```
bdewilde, line 95: I feel strongly that `model` should be an input to evaluation functions, and not loaded within one. For pretty much all loading/saving of data, we have the `dataio` subpackage. in fact, we have a dedicated function for this very thing! https://github.com/datakind/student-success-tool/blob/develop/src/student_success_tool/dataio/models.py#L17-L43

src/student_success_tool/modeling/evaluation.py
```diff
@@ -64,6 +65,144 @@ def extract_training_data_from_model(
...
+    df_pred = df.assign(
+        **{
+            pred_col: model.predict(df_features),
+            pred_prob_col: model.predict_proba(df_features)[:, 1],
+        }
+    )
```
bdewilde, line 101: We also have a function for this: https://github.com/datakind/student-success-tool/blob/develop/src/student_success_tool/modeling/inference.py#L14-L61

The logic here makes assumptions about `pos_label` that aren't made in other functions, so could easily get out of alignment if something were to change. Better to use the general-purpose existing function.

src/student_success_tool/modeling/evaluation.py
```diff
@@ -64,6 +65,144 @@ def extract_training_data_from_model(
     return df_loaded
 
 
+def load_model_and_predict(
```
bdewilde, line 68: stepping back, i think we should probably get rid of this function, and just use existing functions to load a model, make predictions, and then compare models by performance.

src/student_success_tool/modeling/evaluation.py
```diff
@@ -64,6 +65,144 @@ def extract_training_data_from_model(
...
+
+    # Sort and select top run IDs based on min/max with loss vs. score
+    ascending_order = optimization_metric == "log_loss"
+    top_run_ids: t.List[str] = (
+        runs.sort_values(by=search_metric, ascending=ascending_order)
+        .iloc[:topn_runs_included]["run_id"]
+        .tolist()
+    )
+
+    return top_run_ids
```
bdewilde, line 203: nice, this function is very useful 👍

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -1,7 +1,12 @@
+import mlflow
+import logging
 import pandas as pd
 import numpy as np
+import matplotlib.figure
+import matplotlib.pyplot as plt
+import seaborn as sns
 import scipy.stats as st
-from sklearn.metrics import confusion_matrix
+import sklearn.metrics
```
bdewilde, line 9: good!

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -1,7 +1,12 @@
+import mlflow
+import logging
```
bdewilde, line 2:
```suggestion
import logging

import mlflow
```

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -20,17 +25,278 @@
...
+                    pos_label=pos_label,
+                    zero_division=np.nan,
+                ),
+                2,
+            ),
+            "Log Loss": round(
+                sklearn.metrics.log_loss(labels, pred_probs, labels=[False, True]),
+                2,
+            ),
+        }
```
bdewilde, line 185: I recently re-implemented much of this functionality in `evaluation.py` -- [here](https://github.com/datakind/student-success-tool/blob/develop/src/student_success_tool/modeling/evaluation.py#L67-L115) -- and i made several improvements to the logic. Could you just call that function directly? then you'd have something like

```python
eval_metrics = evaluation.compute_classification_eval_metrics(...)
subgroup_metrics = {...}
metrics = eval_metrics | subgroup_metrics
```

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -154,99 +420,120 @@ def generate_bias_flag(
...
 
-    return bias_flags
+    ax.set_yticks(y_positions)
+    ax.set_yticklabels(subgroups)
+    ax.set_ylim(-1, len(subgroups))
+    ax.set_ylabel("Subgroup")
+    ax.set_xlabel("False Negative Parity Rate")
+    ax.set_title(
+        f"""FNPR @ 0.5 for {fnpr_data[0]["group"]} on {fnpr_data[0]["split_name"]}"""
+    )
```
bdewilde, line 534: I'm not 100% sure, but I think you can combine many of these calls using something like

```python
ax.set(yticks=y_positions, yticklabels=subgroups, ylim=(-1, len(subgroups)), ylabel="subgroup", xlabel="false negative parity rate", title="TITLE")
```

worth checking, just to tidy this up! https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set.html
```

File: src/student_success_tool/modeling/bias_detection.py [hunk 9]
```diff
@@ -20,17 +25,278 @@
...
+def evaluate_bias(
+    run_id: str,
+    df_pred: pd.DataFrame,
+    split_col: str,
+    student_group_cols: list,
+    target_col: str,
+    pred_col: str,
+    pred_prob_col: str,
+    pos_label: str,
+) -> None:
```
bdewilde, line 43: ah, I see that you're deliberately mirroring the API for `evaluate_performance()`. that's a nice idea! my comments over there also apply here, so I encourage you to tidy them up in the same ways :)

bdewilde (changes requested):
I saw that you made some changes in the bias-detection code, in addition to the refactor, but I got a bit bogged down. Here's a first pass on suggestions. This is definitely moving in a good direction, but needs a bit more work 🙏

src/student_success_tool/modeling/evaluation.py
vishpillai123: I'm fine with that, but I'm not finding where the make predictions or load model functions are. Do they exist in our current repo? If so, can you point me to them? I'm not seeing them in modeling.utils.

bdewilde: I linked to them in preceding comments:

- https://github.com/datakind/student-success-tool/blob/develop/src/student_success_tool/dataio/models.py#L17-L43
- https://github.com/datakind/student-success-tool/blob/develop/src/student_success_tool/modeling/inference.py#L14-L61

src/student_success_tool/modeling/bias_detection.py
vishpillai123: Okay I can write perf metrics. Though, the formatting is different. This is the table that is "external" and we will be sharing with institutions so I'm going to edit the formatting in the original function if that's okay.

vishpillai123:
hi @bdewilde changes have been made, so please re-review when you can. I need to move on to other priorities, so if there's anything else you want to tidy up, I ask if we can merge this PR and move those changes into the next one (if you want to create one, I'll review!).

BTW, the code is a bit slower for each run because of the extra processing for each subgroup (rather than computing eval and FNPR together, we are calling your eval function then aggregating, etc.) Previous code that took 7 mins for 5 runs now takes 9 mins. We can work with this for now especially since it's much more organized than before, but optimizing this code for performance is preferable long-term.

src/student_success_tool/modeling/evaluation.py
```diff
@@ -106,10 +200,7 @@ def compute_classification_eval_metrics(
         "recall": recall,
         "f1_score": f1_score,
         "log_loss": sklearn.metrics.log_loss(
-            targets, pred_probs, sample_weight=sample_weights
-        ),
-        "roc_auc": sklearn.metrics.roc_auc_score(
-            targets, pred_probs, sample_weight=sample_weights
+            targets, pred_probs, sample_weight=sample_weights, labels=[False, True]
```
bdewilde, line 203: trying to figure out why you added `labels` here?

vishpillai123: If you try to run the code when there’s only one class (y_true = 1), the log loss function throws a ValueError otherwise.

bdewilde (approved):
full disclosure, i don't have a full grasp of everything you've done in `bias_detection`, but i trust that it works correctly. we can indeed work on speed later.

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -101,7 +101,7 @@ def evaluate_bias(
                 plt.close()
 
                 for flag in group_flags:
-                    LOGGER.info(
+                    LOGGER.warning(
```
bdewilde, line 104: makes sense! but i should ask: why aren't info-level messages showing up in your console? (assuming by "console" you mean "databricks notebook".)

vishpillai123: I'm not sure why honestly. I'm seeing this still. Perhaps I'm disabling the logger somehow?
image

vishpillai123: yeah, but it's not showing the logger warnings or info messages that are inside evaluation.py or bias_detection.py.

PR #131, opened 2025-04-11
Add binary course subject in term features

kaylawilding:
## changes
Added a few new feature types : 
1. We already included counts of how many types of courses (key subject areas, etc.) we took, but added in now a binary of if a course type was taken during the term. (Note, this isn't new, but I did test it out and we can include lists of these to combine them, e.g. instead of just flagging math we can flag math or english taken. So, I did not need to add this functionality.)
2. Analogously, we included cumulative counts of types of courses taken, but I added in a binary of this type was taken up to this point. 
3.  Finally added a functionality to be able to flag if a course type was taken within a certain number of credits (12 credits as the default for right now, but we can expand this to make it flexible if people desire). 

## context
This adds in custom features that were used and selected in models for the current pdp schools. These are features requested by schools.  I needed to add these in so that we can re-train and get the pipeline working for the pdp schools.


src/student_success_tool/features/pdp/cumulative.py
```diff
@@ -44,6 +51,7 @@ def add_features(
                 ("student_pass_rate_above_sections_avg", "sum"),
                 ("student_completion_rate_above_sections_avg", "sum"),
             ],
+            credits=12,
```
vishpillai123, line 54: should we add this as an input parameter?

src/student_success_tool/features/pdp/constants.py
```diff
@@ -14,3 +14,4 @@
 
 NUM_COURSE_FEATURE_COL_PREFIX = "num_courses"
 FRAC_COURSE_FEATURE_COL_PREFIX = "frac_courses"
+DUMMY_COURSE_FEATURE_COL_PREFIX = "took_course"
```
vishpillai123, line 17: So, does this look at every course and check whether it was taken in a particular term? This would generate loads of features right?

src/student_success_tool/features/pdp/cumulative.py
```diff
@@ -91,6 +100,8 @@ def expanding_agg_features(
     *,
     num_course_cols: list[str],
     col_aggs: list[tuple[str, str | list[str]]],
+    credits: int,
```
vishpillai123, line 103: Can we get some more information on this input? Is this a credit limit or threshold? You can specify in the docstring.

src/student_success_tool/features/pdp/cumulative.py
bdewilde: I believe we already have this as an input parameter -- see `min_num_credits_full_time` -- and yes, we should use the parameter rather than a hard-coded value.

src/student_success_tool/features/pdp/student_term.py
```diff
@@ -458,6 +468,44 @@ def sum_dummy_cols_by_group(
...
+            f"{col}_{'|'.join(str(item) for item in val)}"
+            if isinstance(val, list)
+            else f"{col}_{val}"
+        )
+        temp_col_series[temp_col] = shared.compute_values_equal(df[col], val)
+    return (
+        df.assign(**temp_col_series)
+        .reindex(columns=grp_cols + list(temp_col_series.keys()))
+        .groupby(by=grp_cols, observed=True, as_index=True)
+        .agg("max")
```
bdewilde, line 499: Why are we aggregating with a "max" ?

src/student_success_tool/features/pdp/student_term.py
```diff
@@ -458,6 +468,44 @@ def sum_dummy_cols_by_group(
...
+        .rename(columns=_rename_dummy_by_group_col)
+        .reset_index(drop=False)
+    )
+
+
+def _rename_dummy_by_group_col(col: str) -> str:
+    return f"{constants.DUMMY_COURSE_FEATURE_COL_PREFIX}_{col}"
```
bdewilde, line 506: I'm probably missing something, but why not just generate these column names directly, rather than "renaming" afterwards? We generate the original column names, right?

src/student_success_tool/features/pdp/student_term.py
```diff
@@ -458,6 +468,44 @@ def sum_dummy_cols_by_group(
...
 
+def equal_cols_by_group(
+    df: pd.DataFrame,
+    *,
+    grp_cols: list[str],
+    agg_col_vals: list[tuple[str, t.Any]],
+) -> pd.DataFrame:
+    """
+    Compute equal to specified values for all ``agg_col_vals`` in ``df``,
+    then group by ``grp_cols`` and aggregate with a "max".
```
bdewilde, line 479: Just checking my understanding: This is similar to the features that compute the _number_ of courses taken with a given attribute, only it's a boolean value? If that's right, why not just leverage the number features, since there's (marginally) more info there? Or, alternatively, is there a way to "consolidate" the information to reduce redundancy of features?

src/student_success_tool/features/pdp/cumulative.py
kaylawilding: Yep, added context

src/student_success_tool/features/pdp/cumulative.py
kaylawilding: Yep, edited to be a param

src/student_success_tool/features/pdp/constants.py
kaylawilding: No, this will only look at the specified subject areas/courses in the config same as we do now for the number of these courses.

src/student_success_tool/features/pdp/student_term.py
kaylawilding: Just aggregating the binary (at the course level) up to another binary (at the student term level) so a max should work here

src/student_success_tool/features/pdp/student_term.py
kaylawilding: I wanted to keep it consistent with the numeric features for ease of understandability, but if you think its a big loss in efficiency, etc., we can edit.

src/student_success_tool/features/pdp/student_term.py
bdewilde: If these features are only intended as binary versions of the corresponding "cumnum" integer features, should we not just piggy-back on those features directly? i.e. `df[some_feature_cumnum].gt(0)`

src/student_success_tool/features/pdp/student_term.py
bdewilde: Yeahh I think I buy that argument, but I'm still wondering if there's a way to consolidate the features _conceptually_ and/or technically. I mentioned one possibility above (`df[some_feature_cumnum].gt(0)` => a boolean version of the feature), but the current approach feels like a missed opportunity for something sharper. Sorry, I know that's not very specific...

src/student_success_tool/features/pdp/constants.py
```diff
@@ -1,5 +1,6 @@
 DEFAULT_MIN_PASSING_GRADE = 1.0
 DEFAULT_MIN_NUM_CREDITS_FULL_TIME = 12.0
+DEFAULT_COURSE_CREDIT_CHECK = 12
```
bdewilde, line 3: So the "course credit check" _isn't_ just another name for the number of credits designating "full-time" enrollment? When I saw `credits=12` I instantly assumed it was meant to flag course enrollments while full-time enrolled.

src/student_success_tool/features/pdp/student_term.py
bdewilde: Thinking a bit more... This might dovetail nicely with a feature (family) I suggested in an Asana ticket to @nm3224 : The (avg) course grade in X key course and/or Y key subject area. Maybe doing grade and boolean "has-taken" features for the key course features would be, together, better than the current cumnum enrollment features we have now. What do you think? (I think key subject areas lend themselves better to cumnum than "has-taken" forms, but the avg grade would be good in this case as well.)

src/student_success_tool/features/pdp/constants.py
kaylawilding: No, it is not, that is why I added a separate variable here and added a note in the docstring explaining.

src/student_success_tool/features/pdp/student_term.py
kaylawilding: Hm, have thought about it, and we could do that. I am thinking though, since 1. mostly these features are specifically requested and 2. it could be helpful to not have binary and numeric, is that I would like to add in the future the ability to specify in the config input whether you want the numeric or binary. In that case, I could be wrong, but it may be better to create them separately, rather than creating both and dropping the numeric in the case where you wanted the binary? What do you think?

bdewilde: I'm like 7/10 on thinking that sounds overly complicated... We should always be down for improving our (standard PDP) feature sets, but practically speaking, we can't and shouldn't say "yes" to every feature request. If you came back and said "I did [this related thing] instead", and got good results, then that to me feels like a best possible outcome.

bdewilde: ha, sorry. i'm saying that supporting either via configuration feels over-complicated, and we'd be better served doing one or the other. that said, i proposed an alternative in some other comment about how we might revise these features more holistically...

src/student_success_tool/features/pdp/student_term.py
kaylawilding: I think we could add that/leave that in the task about possible new features, but it's out of scope for this PR. This PR is focused on adding features people were already using to get us to a standardized set so we could use the inference pipeline for all current PDP schools.

src/student_success_tool/features/pdp/student_term.py
kaylawilding: I suppose since people can for include these features, I suppose it could be redundant, and I agree it could be complicated so I can concede on that. I changed the code though as you suggested for the binary features, but I stick with they should be included. They are only added if someone has specified these courses or subjects to look at and for each of those it really only includes an additional 2-3 features.

vishpillai123 (approved):

PR #132, opened 2025-04-11
trying to add a num_days_since_enrolled feature

nm3224:
## changes
Trying to add a `num_days_since_first_enrolled` feature using: (Term End Date - Cohort Begin Date)

## context
- We don't have a `term_end_date` feature, but I made one. 
- Alternatively, could calculate using (Term End Date - min(Term Begin Date))...  but idk how to do the groupby for students first and not calculate it at the student-term level for each row. (I attempted something which I commented out for now if we want to just use the `cohort_start_dt` instead.) 
- Some schools have had misaligned cohort start dates and first term dates, which makes me think using `term_start_dt` might be better- but using `cohort_start_dt` would be easiest since that's consistent across rows.
- I have something commented for potential use with `term_start_dt`

## questions
Any advice/suggestions?

### DIFF
--- a/src/student_success_tool/features/pdp/cumulative.py
+++ b/src/student_success_tool/features/pdp/cumulative.py
@@ -273,8 +273,6 @@ def _compute_cumfrac_terms_enrolled(
     return cumfrac_terms_enrolled.astype("Float32")
 
 
-
-
 #######################
 # DEPRECATED / BACKUP #
 # only dig into these if databricks starts freaking out about expanding aggs func above
--- a/src/student_success_tool/features/pdp/cumulative.py
+++ b/src/student_success_tool/features/pdp/cumulative.py
@@ -273,6 +273,8 @@ def _compute_cumfrac_terms_enrolled(
     return cumfrac_terms_enrolled.astype("Float32")
 
 
+
+
 #######################
 # DEPRECATED / BACKUP #
 # only dig into these if databricks starts freaking out about expanding aggs func above
--- a/src/student_success_tool/features/pdp/student_term.py
+++ b/src/student_success_tool/features/pdp/student_term.py
@@ -228,32 +228,32 @@ def year_of_enrollment_at_cohort_inst(
     return pd.Series(np.ceil((dts_diff + 1) / 365.25), dtype="Int8")
 
 
-def num_days_since_first_enrolled(  
-    df: pd.DataFrame,  
-    *,    
-    # using cohort_start_dt 
+def num_days_since_first_enrolled(
+    df: pd.DataFrame,
+    *,
+    # using cohort_start_dt
     cohort_start_dt_col: str = "cohort_start_dt",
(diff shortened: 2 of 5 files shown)


bdewilde:
What's the use case / intention behind this feature that the "year of enrollment" feature doesn't cover?

nm3224:
> What's the use case / intention behind this feature that the "year of enrollment" feature doesn't cover?

@bdewilde Honestly, I'm not entirely sure- I'm going down a list of suggested features to add on this task list that I think Kelsey added a while ago (https://app.asana.com/1/6325821815997/project/1208742287493159/task/1209774405734557?focus=true)
I feel like days since enrollment could be richer in information than just the year? We can also see it on a term-level, whereas year of enrollment is the whole year? I know we have a term-rank feature as well but this seems different.. we can ask at the team meeting if you're strongly opposed. I don't see a downside to adding this and letting the feature selection process decide which is more valuable to the model.

nm3224:
> I agree that "year of enrollment" would be probably cover the use cases here and it's not clear how much value this feature would bring. I do think it's interesting to get a # of days vs. # of years (how we have it now), but again it's not clear how useful it might be.
> 
> I'm thinking about it from the standpoint of "year of enrollment" can be useful when students leave and then come back years later... or if they take a very long time to finish. Years probably covers this sufficiently, and we don't need to look at days. Does this make sense? How are you thinking about it? @nm3224
> 
> As far as Kelsey's list, I think it could be helpful to see what functionality we have in comparison to see what we already have (and can de-prioritize) vs. what we don't have but wouldn't be as useful (also maybe deprioritize) vs. what we don't have and would be useful (higher priority).

I feel like days since enrollment could be richer in information than just the year? We can also see it on a term-level, whereas year of enrollment is the whole year? I know we have a term-rank feature as well but this seems different... I also don't see a downside to adding this and letting the feature selection process decide which feature (num days vs year of enrollment... or both) is more valuable to the model. I agree though, I think we should discuss at the team meeting what we do/don't want to add and then I can continue working on that!

bdewilde:
The issue is that "days since enrollment" is, in practice, discretized to "terms since enrollment", since we can't timestamp at finer granularity than the term level. Does "terms since enrollment" have more information than "years since enrollment"? Just a bit, yes! Maybe that's worth including, and letting the model decide.

vishpillai123:
I agree that a term feature may provide slightly more information than just years, but I also think it’s important to not feed both features that we know to be covariant into our feature selection and modeling process. 

If somehow feature selection doesn’t catch one of them, our model will likely choose one of them but the two features will likely have interactions with each other (which can make interpretability confusing and may affect performance).

I think it’s better to simplify our feature generation process as much as possible, especially when we know we can derive one feature from another (we can derive years from terms, for example).

bdewilde:
Just to be clear, the years column is needed to compute targets, so we can't get rid of it. :)

PR #136, opened 2025-04-12
merge v0.2.0 release changes

bdewilde:
## changes

includes all changes from `develop` and feature branches since the v0.1.1 release, constituting the v0.2.0 release of the package.

## context

see changelog entry


bdewilde:
Oops, I did this backwards: You have to merge `main` into `develop`?

vishpillai123 (approved):
Approving so we're not blocked but just a few small things I flagged in the comment above.

bdewilde:
> Everything in SRC looks good to me. Just double checking the changes outside of the SRC folder, I saw there's changes in the .github folder (action.yml, style.yml, type-check etc.) and uv.lock. this is okay right? 
> Also, how about the pandas and pandera changes in pyproject.toml? I'm sure that this was all thought out, but just double checking.

yup! this is intentional. we made changes to our CI workflows, and we also made changes to the dependencies in this package.

PR #138, opened 2025-04-16
renaming FNPR to FNR for technical accuracy

vishpillai123:
## changes
- Changed all naming of FNPR -> FNR.
- Clarified documentation and added references.

## context
- The term FNPR or False Negative Parity Rate (what I've been using) is not commonly used and technically incorrect. We are actually looking at False Negative Rate Parity or FNR Parity, and specifically looking at the absolute percentage differences in FNR. 
- Parity can be either analyzed via a ratio with `FNR_subgroup1 / FNR_subgroup2`, or absolute percentage differences in FNR so `abs(FNR_subgroup1 - FNR_subgroup2)`. 
- This may seem like semantics (and it is), but we want to be accurate when reporting to institutions and in our model cards.
- Another consideration is why don't we look at FNR ratios instead of FNR differences when analyzing parity? 
  - As an example, if we used a ratio and established a threshold of 1.5 (typical standard for flagging), then 30% and 20% or 3% and 2% will both be flagged. 
  - Thus, thresholds with FNR ratios can be less reliable when we have small FNRs. 
  - Honestly, our FNRs can range anywhere from 0 to 50% (maybe even higher), but usually are in the 15% to 40% range. 
  - FNR differences are also more interpretable and we can perform hypothesis testing more easily than with ratios.
- Future work: we will later evaluate our thresholds for FNR differences in May when we do our refactoring. 

## questions
- Does my thought process for this make sense? 
- References:
  - https://docs.oracle.com/en-us/iaas/tools/automlx/latest/legacy/v23.2.3/fairness.html (allows for input of "diff" or "ratio" for parity)
  - https://fidelity.github.io/jurity/about_fairness.html
  - https://afraenkel.github.io/fairness-book/content/05-parity-measures.html#:~:text=A%20false%20positive%20is%20an,graduate%20rate%20of%20admitted%20students.
  - https://search.r-project.org/CRAN/refmans/fairness/html/fnr_parity.html

### DIFF
--- a/src/student_success_tool/modeling/bias_detection.py
+++ b/src/student_success_tool/modeling/bias_detection.py
@@ -17,7 +17,7 @@ LOGGER = logging.getLogger(__name__)
 # Z-score for 95% confidence interval
 Z = st.norm.ppf(1 - (1 - 0.95) / 2)
 
-# Flag FNPR difference thresholds
+# Flag FNR difference thresholds
 HIGH_FLAG_THRESHOLD = 0.15
 MODERATE_FLAG_THRESHOLD = 0.1
 LOW_FLAG_THRESHOLD = 0.05
@@ -52,7 +52,7 @@ def evaluate_bias(
 ) -> None:
     """
     Evaluates the bias in a model's predictions across different student groups for a split
-    denoted by "split_name" across df_pred. For each student group, FNPR (False Negative Positive Rate) is
+    denoted by "split_name" across df_pred. For each student group, FNR (False Negative Rate) is
     computed and any detected biases are flagged. Then, the metrics & plots are logged to MLflow.
 
     Args:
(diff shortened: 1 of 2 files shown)


src/student_success_tool/modeling/bias_detection.py
```diff
@@ -66,12 +67,17 @@ def evaluate_bias(
         pred_prob_col: Column name for the model's predicted probabilities
         pos_label: Label representing the positive class
         sample_weight_col: Column name representing sample weights for the model.
+
+    References:
```
bdewilde, line 71: nice

bdewilde (commented):
I think this all makes sense. I didn't see any FNPR remnants, but I didn't do a super thorough check. Maybe someone else wants to do a deeper dive?

vishpillai123:
> I think this all makes sense. I didn't see any FNPR remnants, but I didn't do a super thorough check. Maybe someone else wants to do a deeper dive?

Thanks Burton! Can someone else help double check this that there are no "FNPR remnants" before we merge? Should be like a 5 minute check tops. @kaylawilding @nm3224 @Mesh-ach

kaylawilding (approved):

PR #139, opened 2025-04-17
add readme instructions

bdewilde:


bdewilde:
hmm, that merge commit shouldn't be in here! going to start a fresh branch and re-submit

PR #142, opened 2025-04-18
removing sla plots from evaluation

vishpillai123:
## changes
- Removed SLA plots and analysis from evaluation.
 
## context
- We discussed as a team, and it appears that the team doesn't find it very useful for evaluating a model's performance.

## questions
- Did I miss anything? Looks like it's all covered in the evaluation file.


bdewilde (approved):
looks good to me!

PR #143, opened 2025-04-22
[pdp] improve project config structure

bdewilde:
## changes

- separates selection and checkpoint configuration into their own sections under "preprocessing" in the project config, which aligns more closely with code structure and actual usage
- updates template project config and unit tests to match
- adds a new shared type for "intensity time limits" that is used across various parts of the code, including the updated project config

## context

this is an awkward part of the code that i'd wanted to fix before v0.2.0, but it didn't get there in time

### DIFF
--- a/pipelines/institution_id/config-TEMPLATE.toml
+++ b/pipelines/institution_id/config-TEMPLATE.toml
@@ -41,9 +41,12 @@ key_course_ids = ["ENGL101", "MATH101"]
 student_criteria = { enrollment_type = "FIRST-TIME", credential_type_sought_year_1 = "Bachelor's Degree" }
 intensity_time_limits = { FULL-TIME = [3.0, "year"], PART-TIME = [6.0, "year"] }
 
-[preprocessing.checkpoint.params]
-min_num_credits = 30.0
-num_credits_col = "num_credits_earned_cumsum"
+[preprocessing.checkpoint]
+name = "30_credits_ckpt"
+params = { min_num_credits = 30.0, num_credits_col = "num_credits_earned_cumsum" }
+
+[preprocessing.target]
+name = "graduation_150pct_time"
 
 [preprocessing.target.params]
 intensity_time_limits = { FULL-TIME = [3.0, "year"], PART-TIME = [6.0, "year"] }
--- a/pipelines/institution_id/config-TEMPLATE.toml
+++ b/pipelines/institution_id/config-TEMPLATE.toml
@@ -37,9 +37,16 @@ core_terms = ["FALL", "SPRING"]
 key_course_subject_areas = ["24", "51"]
 key_course_ids = ["ENGL101", "MATH101"]
 
-[preprocessing.target]
-params = { min_num_credits_checkin = 30.0, min_num_credits_target = 60.0 }
+[preprocessing.selection]
 student_criteria = { enrollment_type = "FIRST-TIME", credential_type_sought_year_1 = "Bachelor's Degree" }
+intensity_time_limits = { FULL-TIME = [3.0, "year"], PART-TIME = [6.0, "year"] }
+
+[preprocessing.checkpoint.params]
+min_num_credits = 30.0
+num_credits_col = "num_credits_earned_cumsum"
+
+[preprocessing.target.params]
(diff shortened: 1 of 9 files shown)


pipelines/institution_id/config-TEMPLATE.toml
```diff
@@ -37,9 +37,16 @@ core_terms = ["FALL", "SPRING"]
 key_course_subject_areas = ["24", "51"]
 key_course_ids = ["ENGL101", "MATH101"]
 
-[preprocessing.target]
-params = { min_num_credits_checkin = 30.0, min_num_credits_target = 60.0 }
+[preprocessing.selection]
 student_criteria = { enrollment_type = "FIRST-TIME", credential_type_sought_year_1 = "Bachelor's Degree" }
+intensity_time_limits = { FULL-TIME = [3.0, "year"], PART-TIME = [6.0, "year"] }
```
vishpillai123, line 42: `intensity_time_limits` in both selection and target params? Why specify it twice?

vishpillai123 (commented):
Quick feedback/question below. 

I was also curious if the outcome variable/target can be more explicitly defined in the config? This is probably out-of-scope for this particular PR. We need to extract the outcome from somewhere for the model cards and I think calling ALL possible targets "target" isn't informative.

vishpillai123:
> Quick feedback/question below.
> 
> I was also curious if the outcome variable/target can be more explicitly defined in the config? This is probably out-of-scope for this particular PR. We need to extract the outcome from somewhere for the model cards and I think calling ALL possible targets "target" isn't informative.

Basically, I'm trying to see if there a way for us to specify all the inputs of the model_name in the config so that it's more programmatic. I know we have this code to create the model_name. But can we specify the target and checkpoint here in the config, as well?

```
model_name = modeling.registration.get_model_name(
        institution_id = cfg.institution_id,
        target = "graduation",
        checkpoint = "2_terms",
        )
```
So we specify the target and checkpoint, as well in the config.

pipelines/institution_id/config-TEMPLATE.toml
bdewilde: For certain targets, it's used in two separate operations: One, to determine which students are eligible for having targets computed, and two, to determine the values of the target itself. I've gone back and forth about whether it makes sense to consolidate the logic, such that the target funcs would output NULL or something in case the dataset doesn't contain enough data to accurately compute the target. But it was a bigger refactor than I've had time for. That said, these changes are meant to better reflect the current code, not change it.

bdewilde: I'll also add that data preprocessing / modeling tends to have lots of parameters, and it's not uncommon that the same parameters get used in different parts of the pipeline. It would be great if this format allowed for sharing parameters, but alas, not a TOML feature. I know of systems that _do_ facilitate this pattern, but it would entail a significant chunk of work and buy-in. So, not going that route unless it becomes critical.

vishpillai123: I see, that makes sense that it’s reflecting the code as is. My concern though is that if this variable should be the same across both functions, then we should define it just once. Otherwise, it’s confusing as a developer and mistakes could also be made if there are accidentally differences in how it’s defined. 

Is it possible for us to define it in more of a generic part of the config so that we can access it from both functions?

bdewilde:
> Basically, I'm trying to see if there a way for us to specify all the inputs of the model_name in the config so that it's more programmatic. I know we have this code to create the model_name. But can we specify the target and checkpoint here in the config, as well?

@vishpillai123 good idea :)

```python
model_name = modeling.registration.get_model_name(
    institution_id = cfg.institution_id,
    target = cfg.preprocessing.target.name,
    checkpoint = cfg.preprocessing.checkpoint.name,
)
```

vishpillai123 (approved):

pipelines/institution_id/config-TEMPLATE.toml
bdewilde: If you're curious about how things could look using a different tool, I'd direct you to `confection`: https://github.com/explosion/confection/

I've used it in other projects and found it convenient and useful (although sometimes finicky/constraining). However, it's only _minimally_ maintained, and not currently compatible with pydantic v2 -- so, incompatible with our current package., womp.

PR #144, opened 2025-04-25
refactor pkg structure

bdewilde:
## changes

- moves `checkpoints`, `features`, `selections`, and `targets` sub-packages _under_ the `preprocessing` sub-package, updates imports accordingly, but otherwise leaves the code as-is
- moves `preprocessing.pdp.dataops` to just `preprocessing.pdp`
- deletes old target modules (hold-overs from v0.1.x, not to be used anymore) and an old readme that was out of date / no longer relevant

this is a proposal to see if you're interested, not something i'm going to push super hard for. i think it's an improvement, but am open to differing viewpoints :)

## context

now that the package's pipeline / components have firmed up, it's a good time to consider package structure more holistically. this code structure intentionally matches the updated project configuration's structure, which should facilitate development / streamline developers' mental models.

## questions

- what do you think of this restructure?
- is there anything you'd like to see added and/or changed here? (note: i'm deliberately limiting the blast radius to just preprocessing -- yes, other bits could also be restructured, but that's out-of-scope.)


vishpillai123 (commented):
I'm curious to hear Kayla's take, but I want to more conservative about when to restructure the package. In all honesty, I don't see this particular restructuring to be worth the refactoring that follows. I think it provides some cleanup/clarity.

vishpillai123:
As far as deleting anything not relevant, I'm on board with that! Just want to be careful about how often we restructure the package since it has clear downstream effects.

bdewilde:
I've been thinking about this more... To me, the 1:1 correspondence with the project config's various blocks is a strong positive signal that this structure makes sense, and will feel more intuitive for developers. It may also lend itself to creating a single, generic `preprocess_data()` function that works for any school, given appropriate inputs.

Beyond this, we should probably tinker with the relationship between `dataio` and `schemas`, possibly splitting out the project config functionality (currently partnered with data schemas, but they're actually different things...) Maybe I'll mock that up to in a follow-up PR? please hold

bdewilde:
pt 2 here: https://github.com/datakind/student-success-tool/pull/146

vishpillai123:
> > I'm curious to hear Kayla's take, but I want to more conservative about when to restructure the package. In all honesty, I don't see this particular restructuring to be worth the refactoring that follows. I think it provides some cleanup/clarity.
> 
> to be clear, the full extent of the developer-facing changes here are tweaks in import paths. shouldn't be disruptive! my stance is, i don't feel the need to be conservative about restructuring _at this point_, since many of the sub-packages i'm moving here just got added in the previous version. imo this sort of thing is normal -- and good! also, since v0.2.0 is released and not missing critical bits of functionality, this sort of change going from v0.2 to v0.3 is "safe" in the sense that folks will only update deliberately, when they want to pull in new features; if not, they can keep using v0.2, nbd.

I see, okay that's fair that it'll just change the import. I do think these changes provide some more clarity. I wouldn't say it's huge, but if it can lead us to eventually having a "preprocess_func" for all schools, that would be amazing.

kaylawilding:
Okay with deleting old unused stuff. 

For the restructure, I don't feel that strongly. I don't feel like this is necessary and possibly not really helpful as I'm not sure the goal should be to get to one final preprocessing function. I think it kind of continues to nests things further which is maybe something we should be moving away from. But I also get that its kind of odd that like features is called in preprocessing/is already functionally nested and not physically a part of it, so I get some appeal in doing that re-org. As far as the costs, we're all about to refactor anyhow and it is just a change in import paths so I agree it is not that major, if this is something the group wants to do. I would lean towards not doing this though.

vishpillai123 (approved):
Approving based on team alignment

PR #146, opened 2025-04-26
[pdp] refactor pkgs (pt2)

bdewilde:
## changes

- moves project config functionality into new `configs` subpackage
- moves `converters`, `schemas` into existing `dataio` subpackage
- updates src/tests imports accordingly
- updates template notebooks' imports accordingly (including refactoring for the preceding PR)

with these changes (plus this in PR #144), the package's overall structure looks like this:

image

## context

follow-up to PR #144

## questions

- check the template nbs for a sense of how much these changes affect devs implementing schools. pretty minor, right?!


vishpillai123:
Ha!! Not much sooner lol. I didn't catch that, thanks Burton!!

nm3224:
> #144

@vishpillai123 Thanks for looping me in, I appreciate it. This definitely makes sense and seems intuitive.

bdewilde:
> @vishpillai123 Thanks for looping me in, I appreciate it. This definitely makes sense and seems intuitive.

@nm3224 Given your comments here and in DM, I think I should check: Do you get notified if you're tagged as a reviewer on a PR? I already tagged you on both of these PRs, which is a standard way of "looping in".

kaylawilding:
I think this PR helps clarify/cleans up a few things from the last one. I am fine with this if we want to move forward with it.

vishpillai123 (approved):
Approving based on team alignment

nm3224:
> > @vishpillai123 Thanks for looping me in, I appreciate it. This definitely makes sense and seems intuitive.
> 
> @nm3224 Given your comments here and in DM, I think I should check: Do you get notified if you're tagged as a reviewer on a PR? I already tagged you on both of these PRs, which is a standard way of "looping in".

@bdewilde i do! But this is partly my fault in that my DK inbox gets super clunky with a lot of spam + asana emails + etc. I need to clean it out and find a way to prioritize the ones from Git so I see them! Generally- I’m not too nit picky about seeing every PR, just need to be more on top of glossing over the ones that make bigger changes or things get moved around significantly

bdewilde:
> @bdewilde i do! But this is partly my fault in that my DK inbox gets super clunky with a lot of spam + asana emails + etc. I need to clean it out and find a way to prioritize the ones from Git so I see them! Generally- I’m not too nit picky about seeing every PR, just need to be more on top of glossing over the ones that make bigger changes or things get moved around significantly

Got it, thanks for clarifying! Totally agreed, my DK inbox is _a mess_. I'll try to flag you more specifically next time if there's something notable.

PR #147, opened 2025-05-01
[pdp] handle students' time left under the hood in targets

bdewilde:
## changes

- removes old `targets.pdp.shared` funcs, which should have been removed previously when the old targets were removed
- adds new shared funcs that determine which students can have targets computed
- updates graduation target such that boolean values give true labels suitable for model training, while null values indicate "unlabeled" students suitable for inference
- makes year-to-first-degree call required w/o a default value, since it was super easy to accidentally always compute for bachelor's degrees... this was a bad default

TODO:
- [x] finish updating graduation target tests
- [x] totally re-do shared targets tests
- [x] incorporate equivalent logic into other targets

## context

@vishpillai123 recently asked why we had to specify `intensity_time_limits` twice in the project config... these changes allow us to only specify it once, and also makes for significantly easier and less error-prone target / (un)labeled handling during data preprocessing

## questions

- does this change make sense, conceptually?
- is the implementation correct? (pretty sure yes, but this shit is _tricky_)
- do you support implementing this change for all targets?


src/student_success_tool/preprocessing/targets/pdp/graduation.py
```diff
@@ -91,16 +111,42 @@ def compute_target(
         .assign(target=True)
         .astype({"target": "boolean"})
     )
-    return (
-        # all students not assigned True, now assigned False
-        pd.merge(df_distinct_students, df_target_true, on=student_id_cols, how="left")
-        # fill nulls (i.e. non-True) with False and ensure we have "vanilla" bool values
+    # get all students for which a target label can accurately be computed
+    # i.e. the data in df covers their last "on-time" graduation term
+    df_labelable_students = shared.get_students_with_max_target_term_in_dataset(
```
vishpillai123, line 116: I'm not following conceptually why you're focusing on students who have the max target term per intensity time limits. Max target term to me sounds like you're capturing the most recent students who would meet the checkpoint & intensity time limits, but wouldn't you want to capture all students who meet these conditions? Maybe I'm missing where you're doing that?

bdewilde: @vishpillai123 Let me try to explain in words instead of code comments...

Separate from computing targets -- for both training and inference datasets -- we select the student population by attributes using a function in `preprocessing.selection`. For the training dataset only, we additionally need to exclude students for which the dataset doesn't cover enough time for the desired target to be accurately computed. This functionality currently lives in `preprocessing.selection` (e.g. [here](https://github.com/datakind/student-success-tool/blob/develop/src/student_success_tool/preprocessing/selection/pdp.py#L113-L127)), but it's technically a _different thing_. This `get_students_with_max_target_term_in_dataset()` function checks if the latest ("max") term in which a student can graduate and be considered "on-time" is included in the dataset, and returns the set of students for which that is true. These students are "labelable" i.e. we can compute accurate True/False targets for them; students for which the max term isn't present are "unlabeleable", or in more common terminology, _unlabeled_.

Bundling this logic under the hood in the targets does a couple things: 1. it prevents user error, something that the current setup actively allows (I would know, I've error'd this!), 2. it makes it easier to preprocess data in both train and predict run modes, and 3. it avoids the need to repeat `intensity_time_limits` in multiple places in the config. Does that make sense?

vishpillai123: Yes that makes sense! Thanks for clarifying. This looks good to me then, whenever you're ready for me to review let me know.

bdewilde:
closing in favor of PR #149

PR #148, opened 2025-05-03
[pdp] remove old shared target funcs

bdewilde:
## changes

removes shared target functions that are no longer needed, and fixes a few calls to them that should point to equivalents elsewhere in the code

## context

this functionality should've been removed in PR #144, but I forgot -- there was a lot going on there


bdewilde:
Going to merge this in, since it's just removing old code that should've been removed before...

PR #149, opened 2025-05-03
[pdp] handle "unlabel-able" students in target funcs

bdewilde:
## changes

- adds new shared funcs that determine which students can have targets computed based on the "time left" in the dataset
- updates graduation, retention, and num-credit targets such that boolean values give true labels suitable for model training, while null values indicate "unlabeled" students suitable for inference
- makes year-to-first-degree arg in graduation target _required_, w/o a default value, since it was super easy to accidentally always compute for bachelor's degrees... this was not a bad default, my bad

## context

exact same changes as in PR #147, only with the (unrelated) removal of old code split into its own PR (#148)

for a more detailed explanation of why/what: https://github.com/datakind/student-success-tool/pull/147#discussion_r2070978680

## questions

- does this change make sense, conceptually?
- is the implementation correct? (pretty sure yes, but this shit is tricky)
- the template notebook doesn't actually have a concrete implementation for a target. would you want me to add one in, to demonstrate how this stuff works now?


nm3224:
> * the template notebook doesn't actually have a concrete implementation for a target. would you want me to add one in, to demonstrate how this stuff works now?

@bdewilde yes please, would appreciate this!

src/student_success_tool/preprocessing/targets/pdp/credits_earned.py
```diff
@@ -101,13 +109,33 @@ def compute_target(
         .assign(target=True)
         .astype({"target": "boolean"})
     )
-    return (
-        # all students not assigned True, now assigned False
-        pd.merge(df_distinct_students, df_target_true, on=student_id_cols, how="left")
-        # fill nulls (i.e. non-True) with False and ensure we have "vanilla" bool values
+    # get all students for which a target label can accurately be computed
+    # i.e. the data in df covers their last "on-time" graduation term
```
nm3224, line 113: @bdewilde this implementation is quite strict... which I understand why since we only want to include students who we _definitely_ can compute a target for... but even when using the `time_left` function right now, I feel like some students who exist in early enough cohorts for a 4 year or 6 year timeline get dropped because they maybe skipped terms or something. Again, understand the reason for it b/c we don't want to risk the other issue which is labeling students with a target who we don't have data to do that for, but just something I was thinking about.

This still includes students who enrolled early enough to be on a 4 or 6 year timeline but dropped out, though, right? They'd get labeled "False" ?

bdewilde: Agree that this is strict, but I think _correctly_ so. Do you think it fails to label students who should in fact be labeled?

This target is unusual in that the "time left" check isn't necessarily reckoned by the students' first term of enrollment (unlike graduation and retention). Here, we typically have a checkpoint term that's some intermediate number of credits earned; in such cases, students don't end up in the final labeled dataset because they don't have a suitable checkpoint term from which to make predictions. Do you think we shouldn't allow that type of configuration?

bdewilde:
matching updates to the template notebook coming tomorrow(?) -- sorry, i've been splitting time between homestead projects and data work for the last week 😅

src/student_success_tool/preprocessing/targets/pdp/credits_earned.py
nm3224: I think it makes sense, and in my experience it's only been like 5-7 students that should be eligible based on their cohort but get dropped b/c of some issue in the dataset. 

When I use the function, I don't count time from the checkpoint term; I still count from the first term. I noticed when counting from the checkpoint term, students who _shouldn't_ get dropped, are dropped. But in later cells, it's merged with the checkpoint df so that we still only include students who meet the checkpoint. My jefferson notebook code demonstrates this (theyre looking at 24 credit checkpoint and graduation with associate's in 6 years)

src/student_success_tool/preprocessing/targets/pdp/shared.py
```diff
@@ -1,11 +1,159 @@
...
-# LOGGER = logging.getLogger(__name__)
+LOGGER = logging.getLogger(__name__)
+
+
+def get_students_with_max_target_term_in_dataset(
+    df: pd.DataFrame,
+    *,
+    checkpoint: pd.DataFrame | t.Callable[[pd.DataFrame], pd.DataFrame],
+    intensity_time_limits: utils.types.IntensityTimeLimitsType,
+    max_term_rank: int | t.Literal["infer"] = "infer",
```
vishpillai123, line 17: Can we get some more clarity around whether max term implies the oldest or newest term?

src/student_success_tool/preprocessing/targets/pdp/shared.py
```diff
@@ -1,11 +1,159 @@
...
+        if isinstance(checkpoint, pd.DataFrame)
+        else checkpoint(df)
+    )
+    if df_ckpt.groupby(by=student_id_cols).size().gt(1).any():
+        raise ValueError("checkpoint df must include exactly 1 row per student")
+
+    intensity_num_terms = utils.misc.convert_intensity_time_limits(
+        "term", intensity_time_limits, num_terms_in_year=num_terms_in_year
+    )
+    if "*" in intensity_num_terms:
```
vishpillai123, line 69: what is this for? what does "*" represent?

src/student_success_tool/preprocessing/targets/pdp/credits_earned.py
vishpillai123: I agree with Noreen that we shouldn't count from the checkpoint term, but I can't tell if that's what happening here?

It looks like Burton's finding the max term for student's who met the checkpoint. That max term should be ranked not from the checkpoint term (on another note, what is a checkpoint term? lol), BUT it looks like it's from the first term in the dataset or the first cohort term.

vishpillai123 (commented):
@bdewilde  your implementation looks mostly good to me, though I want to ensure we're all on the same page before we move forward. Especially since Noreen will be the one using this 90% of the time.

src/student_success_tool/preprocessing/targets/pdp/shared.py
bdewilde: Read the docs! ;)

> intensity_time_limits: Mapping of enrollment intensity value (e.g. "FULL-TIME")
            to the maximum number of years or terms considered to be "on-time" for
            the target number of credits earned (e.g. [4.0, "year"], [12.0, "term"]),
            where the numeric values are for the time between "checkpoint" and "target"
            terms. Passing special "*" as the only key applies the corresponding time limits
            to all students, regardless of intensity.

src/student_success_tool/preprocessing/targets/pdp/credits_earned.py
bdewilde: I understand that there's some confusion here: Before, this confusion was _out in the open_ for developers to trip over -- I should know, I've tripped on it! -- and these changes are meant to mostly conceal them under the hood, so that mistakes can't be made. For this particular target, we've had cases where the "intensity time limits" were defined from the checkpoint term instead of the first term of enrollment; for that reason, I included the checkpoint as an input parameter for the target.

A "checkpoint" term is, really, just a single term per student that meets a particular set of requirements. We use this functionality to pull out the terms from which model predictions are made; we also use it to determine which students can be accurately labeled or not, based on the time left in the full dataset.

vishpillai123: Got it, I think it is better that this is handling under the hood, but we just need to agree on the exact process. 

 I defer to @nm3224 on whether to move forward on this particular implementation since she's using it for one of her schools.

nm3224: hey all! This makes sense- @bdewilde are you saying it depends on the use case, and some point have the intensity time limits defined from the checkpoint term instead of the first term, but not restricted to that? If so, I think this is okay- I just think for most of my use cases, I will still be counting from the first-term, not the checkpoint

nm3224: (this is also still failing a test + out of date, so once that's done, we can approve and merge!)

nm3224: I also think if it _is_ something that can depend on the use case, having "checkpoint" as the default parameter can get confusing... maybe it would be better to rename? Unless the intent is to always count intensity time limits from the checkpoint term, not cohort start (which for my schools, wouldn't necessarily work)

bdewilde: Hi, once again, the "first term" _is_ a checkpoint. Making it an input parameter allows devs to tailor it to their use case: You could pass students' first terms as a dataframe, or as the `checkpoints.pdp.first_student_terms` function. This gives us the most flexibility while still having strong guardrails about valid inputs.

bdewilde: > (this is also still failing a test + out of date, so once that's done, we can approve and merge!)

I believe it's all up to date and passing tests now 👍

src/student_success_tool/preprocessing/selection/pdp.py
```diff
@@ -66,148 +65,6 @@ def select_students_by_attributes(
     return df_selected
 
 
-def select_students_by_second_year_data(
```
kaylawilding, line 69: Hi, just to be sure. These old selection functions are irrelevant now because they were used for the old targets, correct? Also, I am not seeing the remaining selection function (select_students_by_attibutes) called anywhere nor in the templates. I believe it should be called directly in the template? If yes, no need to hold up this PR - I can add it to a PR Im working on right now.

kaylawilding (approved):

src/student_success_tool/preprocessing/targets/pdp/credits_earned.py
vishpillai123: Given that "checkpoint" can refer to the "first enrolled term" or a "credit checkpoint", I think it might make sense to change this name to something that is more intuitive and doesn't get confused with the credit checkpoint always. I think that's Noreen's request too, so we'll get this merged and then add that as a separate PR.

PR #150, opened 2025-05-05
changing T/F to 1/0

nm3224:
## changes & context
- upon running a model with this new feature, on SHAP values it's still greyed out- which means autoML is NOT changing the T/F to 1/0! I ended up working around this by just doing this in the notebook: df["student_is_pell_recipient_first_year"] = df["student_is_pell_recipient_first_year"].map({True: 1, False: 0}) - i don't specify int b/c this handles the case of <NA> values too (leaves it as NaN) - but wanted to fix here too! Let me know thoughts. 

## questions
any questions?

### DIFF
--- a/src/student_success_tool/preprocessing/features/pdp/student.py
+++ b/src/student_success_tool/preprocessing/features/pdp/student.py
@@ -81,7 +81,7 @@ def student_is_pell_recipient_first_year(
     *,
     pell_col: str = "pell_status_first_year",
 ) -> pd.Series:
-    return df[pell_col].map({"Y": True, "N": False}).astype("boolean")
+    return df[pell_col].map({"Y": 1, "N": 0})
 
 
 def diff_gpa_term_1_to_year_1(


bdewilde:
just curious: did you include this in your process? https://github.com/datakind/student-success-tool/blob/develop/pipelines/institution_id/02-train-model-TEMPLATE.py#L106-L107

vishpillai123 (changes requested):
I think any boolean features should stay boolean and 1/0 variables that we want to treat as boolean should be converted to True/False boolean. 

We don't need any True/False (or 1/0) columns to show up as colored in the SHAP value plot because boolean columns are categorical and shouldn't be treated as numerical. 

I guess I'm not understanding the intention of this. Is this just to show up in the SHAP plot? Just to clarify, boolean and categorical columns will still have SHAP values they just won't be visible in the plot. I'd prefer not to do this as it creates inconsistency across our boolean columns in feature engineering. Are we converting all columns to 1/0 from True/False?

Not to mention, this will create later issues with our output dataframe we give to advisors. We want boolean columns to show up as True/False and not 1/0, otherwise advisors may have trouble understanding the column.

nm3224:
> I think any boolean features should stay boolean and 1/0 variables that we want to treat as boolean should be converted to True/False boolean.
> 
> We don't need any True/False (or 1/0) columns to show up as colored in the SHAP value plot because boolean columns are categorical and shouldn't be treated as numerical.
> 
> I guess I'm not understanding the intention of this. Is this just to show up in the SHAP plot? Just to clarify, boolean and categorical columns will still have SHAP values they'll just be grayed out in the plot. I'd prefer not to do this as it creates inconsistency across our boolean columns in feature engineering. Are we converting all columns to 1/0 from True/False?
> 
> Not to mention, this will create later issues with our output dataframe we give to advisors. We want boolean columns to show up as True/False and not 1/0, otherwise advisors may have trouble understanding the column.

Yes, the intended goal with creating this feature in the first place was so we can have pell status be part of the SHAP values and evaluate it's importance / impact on student success / support needed instead of including it as part of the bias eval (not sure if you were there at the team meeting we talked about this @vishpillai123 , it was a while ago but lmk if it doesn't ring a bell!) We also still have the original column still in the dataframe coded as Y/N so it shouldn't be an issue for advisors imo

nm3224:
> just curious: did you include this in your process? https://github.com/datakind/student-success-tool/blob/develop/pipelines/institution_id/02-train-model-TEMPLATE.py#L106-L107

Yes! Still doesn't change it for whatever reason, not sure if something is missing or what

vishpillai123:
> > I think any boolean features should stay boolean and 1/0 variables that we want to treat as boolean should be converted to True/False boolean.
> > We don't need any True/False (or 1/0) columns to show up as colored in the SHAP value plot because boolean columns are categorical and shouldn't be treated as numerical.
> > I guess I'm not understanding the intention of this. Is this just to show up in the SHAP plot? Just to clarify, boolean and categorical columns will still have SHAP values they'll just be grayed out in the plot. I'd prefer not to do this as it creates inconsistency across our boolean columns in feature engineering. Are we converting all columns to 1/0 from True/False?
> > Not to mention, this will create later issues with our output dataframe we give to advisors. We want boolean columns to show up as True/False and not 1/0, otherwise advisors may have trouble understanding the column.
> 
> Yes, the intended goal with creating this feature in the first place was so we can have pell status be part of the SHAP values and evaluate it's importance / impact on student success / support needed instead of including it as part of the bias eval (not sure if you were there at the team meeting we talked about this @vishpillai123 , it was a while ago but lmk if it doesn't ring a bell!) We also still have the original column still in the dataframe coded as Y/N so it shouldn't be an issue for advisors imo

Yes, I remember we wanted to include it as a feature. Though, I'm still not understanding this PR. Boolean columns can still have importance and SHAP values. All features (numerical or categorical) will have SHAP values. It'll just be grayed out in the plot since there are no numerical feature values, which is fine. I think converting back and forth isn't necessary here!

nm3224:
> > > I think any boolean features should stay boolean and 1/0 variables that we want to treat as boolean should be converted to True/False boolean.
> > > We don't need any True/False (or 1/0) columns to show up as colored in the SHAP value plot because boolean columns are categorical and shouldn't be treated as numerical.
> > > I guess I'm not understanding the intention of this. Is this just to show up in the SHAP plot? Just to clarify, boolean and categorical columns will still have SHAP values they'll just be grayed out in the plot. I'd prefer not to do this as it creates inconsistency across our boolean columns in feature engineering. Are we converting all columns to 1/0 from True/False?
> > > Not to mention, this will create later issues with our output dataframe we give to advisors. We want boolean columns to show up as True/False and not 1/0, otherwise advisors may have trouble understanding the column.
> > 
> > 
> > Yes, the intended goal with creating this feature in the first place was so we can have pell status be part of the SHAP values and evaluate it's importance / impact on student success / support needed instead of including it as part of the bias eval (not sure if you were there at the team meeting we talked about this @vishpillai123 , it was a while ago but lmk if it doesn't ring a bell!) We also still have the original column still in the dataframe coded as Y/N so it shouldn't be an issue for advisors imo
> 
> Yes, I remember we wanted to include it as a feature. Though, I'm still not understanding this PR. Boolean columns can still have importance and SHAP values. All features (numerical or categorical) will have SHAP values. It'll just be grayed out in the plot since there are no numerical feature values, which is fine. I think converting back and forth isn't necessary here!

The idea with having it color coded is so we can actually evaluate how pell status impacts the model output. Otherwise, you just have grey blobs spread across the shap values; but we can't actually interpret how being/not being pell-recipient impacts support needed

vishpillai123:
You still can! It won't show up in the plot, but you can still evaluate by averaging across the output dataframe. You still have the SHAP values.

I understand you did this before, but I think that doing this across all schools isn't necessary and it creates inconsistency with our other boolean features.

nm3224:
> You still can! It won't show up in the plot, but you can still evaluate by averaging across the output dataframe. You still have the SHAP values.
> 
> I understand you did this before, but I think that doing this across all schools isn't necessary and it creates inconsistency with our other boolean features.

Hmm I feel like seeing and explaining it on the SHAP plot is nice visually for schools to see and explain it, especially for those (ex. Rutgers) that are really passionate about specifically supporting pell-recipient students. If we're not going to have it be part of the SHAP values color-coded, then we don't really _need_ this new feature- we can just force include pell status as it stands with Y/N. Maybe something we can revisit at team meeting?

vishpillai123:
> > You still can! It won't show up in the plot, but you can still evaluate by averaging across the output dataframe. You still have the SHAP values.
> > I understand you did this before, but I think that doing this across all schools isn't necessary and it creates inconsistency with our other boolean features.
> 
> Hmm I feel like seeing and explaining it on the SHAP plot is nice visually for schools to see and explain it, especially for those (ex. Rutgers) that are really passionate about specifically supporting pell-recipient students. If we're not going to have it be part of the SHAP values color-coded, then we don't really _need_ this new feature- we can just force include pell status as it stands with Y/N. Maybe something we can revisit at team meeting?

I agree it may be convenient and needed for certain schools, but this shouldn't be reflected in the the feature engineering processing. I think it could be transformed right before you run the SHAP plot right? I don't think we need to integrate this from the feature engineering step. 

Just to clarify, I'm fine with including it as a feature, and yes you can force include it if it's not picked up in feature selection (you do need to take it out of the `student_group_cols`, though). 

I'm very hesitant to convert to 1/0 during feature engineering because it creates inconsistency (and more processing) with our other boolean features.

bdewilde:
I skimmed this thread, and found myself largely agreeing with Vish. If you really want high/low value colors in the SHAP summary plot, you can cast the boolean column(s) to integer(s) immediately prior, and everything should otherwise work correctly. I don't recommend going back to the "Y"/"N" form of this particular feature.

PR #151, opened 2025-05-06
align config to model cards 

vishpillai123:
Our [model cards](https://docs.google.com/document/d/1LXMzOn_IwYjGFNarl33GHUeWGxcUsyg_fCA-pHlxGmA/edit?tab=t.0) require very specific information about targets and checkpoints. Our current configuration setup is too flexible—it defines each config class using only a generic `name` field. To improve clarity and make it easier to generate human-readable descriptions for model cards, I’ve broken this `name` field into its component parts. This change also enables us to build more structured and consistent model names (I’ve also updated the `get_model_name` function to reflect these config changes).

I changed `model_name` to the following format below (need to align with team on this). I utilized 'T' and 'C' to differentiate "target" and "checkpoint" and as shorthand to help trim the name a bit.
`"{institution_id}_{target.category}__[{target.unit}{target.value}]__{checkpoint.unit}{checkpoint.value}_[{checkpoint.optional_desc}]"`

So, this would be an example model name:
`inst01_graduation__4term__90credit`

Also, I also converted `[models.TARGET_CHECKPOINT]` to `[model]` for two reasons:
1. Every config.toml file has a 1-1 mapping to a model. This is from google's recommendation, and I know I previously wanted each config to allow for multiple models, but I think this will create more challenges in standardizing and structuring the config across PDP and custom institutions.
2. TARGET_CHECKPOINT doesn't need to be specified directly, since that information is already captured under `preprocessing.target` and `preprocessing.checkpoint`. This would be more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

## changes
- Removed name and added `unit` and `value` for checkpoint with `optional_desc`. The name can be put together with these components.
- Removed name  and added `category`, `unit`, and `value` for target. The name can be put together with these components.
- Removed `model_name` from the templates for now, or at least when it is specified from the config.
- Added validation such that a unit and value must be specified together for a checkpoint or target.
- Changed how the `model_name` is being put together based on our config changes.

## context
- Provided context above.
- These changes will make model cards, which are an essential part of our business process with SST, more easy to manage.
- Our model names need more standardization as well. Luckily, we can use these new config changes to help us standardize.

## questions
- Do these changes provide enough flexibility with the different types of models we provide for SST?

TODO: Add `model_name` back to templates where it's needed. I think it only really be needed for registration and model cards notebook (which is coming soon).


bdewilde:
@vishpillai123 Much of this is still useful, even after my updates to the project configs, right?? Are you just creating a fresh PR building on top of them, but re-using the logic here?

vishpillai123:
@bdewilde with your updates, I think we’re set for PDP, though we do need to update how we create the model names. That could be a separate PR.

I will likely use this implemention that I have here for custom schools, which is why I’m not deleting the branch yet. Though, we need to first create a project config for custom schools.

PR #152, opened 2025-05-07
[pdp] more and better structured configs

bdewilde:
## changes

- reorders project config file so that classes follow in definition order
- restricts project config to defining a single ("best") model, rather than a mapping of models
- adds separate, fully-specified configs for each target, rather than a generic under-specified target config; also requires a new `type_` field, for model card parsing logic
- forbids extra params in project configs, just to be safe
- updates template config and tests

**TODO:** add checkpoint-specific configs in the same manner as targets

## context

alternative to a subset of changes in PR #151

note: model configs are based on what's implemented in PR #149

## questions

- can this configuration be used to both populate model cards and execute pipelines without duplicate info, and still in a human-friendly form?

### DIFF
--- a/src/student_success_tool/configs/pdp.py
+++ b/src/student_success_tool/configs/pdp.py
@@ -52,13 +52,14 @@ class PDPProjectConfig(pyd.BaseModel):
             "used to load the artifacts from storage"
         ),
     )
-    models: dict[str, "ModelConfig"] = pyd.Field(
-        default={},
+    model: t.Optional["ModelConfig"] = pyd.Field(
+        default=None,
         description=(
-            "Mapping of model name, e.g. 'graduation', to MLFlow metadata used to "
-            "load the trained artifact from storage"
+            "Essential metadata for identifying and loading trained model artifacts, "
+            "as produced by the pipeline represented here in this config"
         ),
     )
+
     # key steps in project pipeline
     preprocessing: t.Optional["PreprocessingConfig"] = None
     modeling: t.Optional["ModelingConfig"] = None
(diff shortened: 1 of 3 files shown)


src/student_success_tool/configs/pdp.py
```diff
@@ -93,43 +246,39 @@ class CheckpointConfig(pyd.BaseModel):
...
-    @pyd.field_validator("splits", mode="after")
-    @classmethod
-    def check_split_fractions(cls, value: dict) -> dict:
-        if (sum_fracs := sum(value.values())) != 1.0:
-            raise pyd.ValidationError(
-                f"split fractions must sum up to 1.0, but input sums up to {sum_fracs}"
-            )
-        return value
+
+class TargetGraduationConfig(TargetBaseConfig):
```
vishpillai123, line 259: I like how we separated this all out for grad, retention, and credits earned. I can adapt this for model cards.

src/student_success_tool/configs/pdp.py
```diff
@@ -93,43 +246,39 @@ class CheckpointConfig(pyd.BaseModel):
     params: dict[str, object] = pyd.Field(default_factory=dict)
```
vishpillai123, line 246: Can we build the same setup for Checkpoint? I would like to avoid the "name" as we have it here and have it more structured like the targets.

vishpillai123 (commented):
This is definitely an improvement. It will clearly only work for PDP, while the solution that I had could in theory work for both (though this is more DRY). I'm fine with rolling with what you have here for now, though I would like us also to standardize the model name from this, as I had in my PR.

We also need the same structure for checkpoint. I would like to avoid the "name".

src/student_success_tool/configs/pdp.py
bdewilde: > TODO: add checkpoint-specific configs in the same manner as targets

that's the idea, i just ran out of time :)

vishpillai123 (approved):
Ok sounds good I can build from this tomorrow

PR #153, opened 2025-05-09
[pdp] structure checkpoint configs

bdewilde:


vishpillai123 (approved):
awesome thanks! we may want to create the name from the other parameters, but we'll work on that in a later PR.

PR #154, opened 2025-05-09
creating template notebook for PDP model cards

vishpillai123:
Added another template notebook that does model registration and model cards. We didn't previously have a notebook that does registration, and model cards fits conceptually since model selection is completed. The exact order of the notebooks (I know I put this 4th) can be changed later and is up for discussion. Regardless, this notebook does need to be added to our process.

I also extended the model cards module to use PDP overrides based on latest config changes (custom is coming soon).

## changes
- Added template notebook that produces a model card markdown. PR #155 will be to convert this markdown into a PDF.
- Extended ModelCard class to PDP (PDPModelCard). This design was for a couple reasons.
   - PDP has a specific config, and inheritance allows us to enforce the config for PDP (custom schools will have a different config).
   - This sets us up nicely to extend this to custom schools (coming soon), as we will create a new subclass called CustomModelCard, or something. 
   - PDP and custom schools will use the same model card template. Though, since the configs will be different, our register_sections within each model card instance is going to be different in certain places (for example: checkpoint, target info).
   - I still need to design a lot of the functionality (project config, model card design etc.) for custom schools and I would like to keep this modular and not affect PDP schools in the process. This design helps us with that.
- Added updates to checkpoint and target sections for PDP based on latest config changes (thanks Burton!)
- I've been testing markdown -> PDF conversion, so there's some formatting changes as well as very small changes to the template. The actual code to do this (along with a .css file) will be added to another PR. All these changes are safe though, so this PR is fine standalone.

## context
- We want to have model cards for custom and PDP schools. They will use the same model card template and share a lot of similarities in the config and model card creation process. Though they also have critical differences and this design enables those differences.
- We can start testing model card markdown for PDP schools.

## questions
- No questions


src/student_success_tool/reporting/platforms/pdp/model_card.py
```diff
@@ -0,0 +1,33 @@
+import typing as t
```
kaylawilding, line 1: I think this should live at the same level as the other model_card.py and maybe rename this to be pdp_model_card.py. I think it would be clearer/more obvious to use this and how to use this.  Also, can you specify in the readme about the model_card.py vs this py file and the different examples?

kaylawilding (commented):
Hi, I think this is really well done. Great modularity, logging, etc, so its very production ready. Just a small thing about where the pdp model_card.py should live, but otherwise great work.

src/student_success_tool/reporting/platforms/pdp/model_card.py
vishpillai123: You wouldn't use the other model card file straight up. It's more of a base for different platforms (pdp vs. custom), so that we share code as much as possible.

The template loads the PDP model card directly so users don't necessarily have to worry about which one to use. I'll add the import line in the README though.

vishpillai123: I can put it in the same area, so that it's less confusing.

vishpillai123:
@kaylawilding hi, I encorporated your feedback. Let me know what you think.

kaylawilding (approved):

PR #155, opened 2025-05-10
add export of markdown to pdf for model cards

vishpillai123:
Followed after PR #154 is merged. For our model cards to be consumed, we need to convert our markdown file into pdf. There's several ways to do this, and I took the markdown -> HTML -> PDF route using CSS. 

For us to do any conversion into PDF, we need to use external packages. For our case, we used weasyprint and markdown, which is why there are new packages in pyproject.toml and uv.lock. 

Page 1 of what the PDF looks like is below. This is a solid start with what we need. Next steps will be looking for feedback from our team, leadership, and the research team (may be a follow-up PR to incorporate their feedback).

## changes
- Added weasyprint and markdown installs/imports
- Added CSS file to be used for all templates. 
- Added editing & conversion in template. After you create your markdown using this model card package, the user can edit the file in markdown, and then convert to PDF. No refreshing of the page or anything is needed, as I wrote functions that will re-read the file whether edits are made to the markdown or not.
- Added log artifact function. This is used just once to log the PDF version of the model card as an artifact. Otherwise, we cannot view the model card in databricks. I'll need to work with Emma and Rachel to see how we want to download and give this to our institutions. 
- Changed image embedding into markdown. We are no longer using set widths, but instead width percentages, which are cleaner and more generalizable. We no longer need to hardcode widths for each image. 
- TODO: Split up evaluation by group tables into separate tables since the width of the tables is too long. 

## context
- PDF versions of our model cards will be needed since model cards are technically reports.
- We will be looking for feedback as this is our first version of the automated template.

Example of Page 1 of rendered model card PDF: 
image

### DIFF
--- a/pipelines/institution_id/00-data-assessment-TEMPLATE.py
+++ b/pipelines/institution_id/00-data-assessment-TEMPLATE.py
@@ -97,8 +97,11 @@ cfg
 # TODO: if dataset info included in cfg, specify its name here
 dataset_name = "DATASET_NAME"
 
-%md
-## course dataset
+
+# COMMAND ----------
+
+# MAGIC %md
+# MAGIC # course dataset
 
 # COMMAND ----------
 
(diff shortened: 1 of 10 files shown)


vishpillai123:
@kaylawilding hi, this is ready for you to review.

kaylawilding (approved):

PR #160, opened 2025-05-14
updating schema for math and english dev and features

nm3224:
NSC shared with us that we should expect PDP data for the completed_dev_english_y_1 and completed_dev_math_y_1 columns not to be populated as "C, D, NA" as expected [here](https://help.studentclearinghouse.org/pdp/knowledge-base/cohort-level-analysis-ready-file-data-dictionary/) . That's a mistake on their documentation. They said we should be expecting inputs as "Y, N, NA". Our current schema validation for the cohort file isn't set up to handle this (expects C, D, NA values as used for all gateway course features) so we need to change that.


vishpillai123:
Shouldn't we change the field and not assign it to a generic field?
```
CompletedDevOrGatewayField = ft.partial(
    pda.Field, nullable=True, dtype_kwargs={"categories": ["C", "D", "NA"]}
)
```

nm3224:
> Shouldn't we change the field and not assign it to a generic pandera field?
> 
> We change this:
> 
> ```
> CompletedDevOrGatewayField = ft.partial(
>     pda.Field, nullable=True, dtype_kwargs={"categories": ["C", "D", "NA"]}
> )
> ```
> 
> to this:
> 
> ```
> CompletedDevOrGatewayField = ft.partial(
>     pda.Field, nullable=True, dtype_kwargs={"categories": ["Y", "N", "NA"]}
> )
> ```

We can change it to that instead of a generic for the specific features (idk how much it matters b/c the attempted dev fields are using generics so I followed that structure) but we shouldn't change the field as a whole b/c that function is still applicable to the gateway courses, just not for the dev courses

src/student_success_tool/dataio/schemas/pdp/raw_cohort.py
```diff
@@ -107,12 +107,8 @@ class RawPDPCohortDataSchema(pda.DataFrameModel):
     attempted_dev_math_y_1: pt.Series[pd.CategoricalDtype] = pda.Field(nullable=True)
```
nm3224, line 107: @vishpillai123 these are generics so I changed it to that

nm3224:
Maybe we change the name of that func tho to be specific to just gateway courses tho @vishpillai123 ?

vishpillai123:
> Maybe we change the name of that func tho to be specific to just gateway courses tho @vishpillai123 ?

Yes good call, we need to separate out `CompletedDevField` and `CompletedGatewayField`. Can we do that instead of that generic field? I know attempted field has that but I would prefer to not remove the validation that we had before, just adjust it.

nm3224:
> > Maybe we change the name of that func tho to be specific to just gateway courses tho @vishpillai123 ?
> 
> Yes good call, we need to separate out `CompletedDevOrGatewayField` into `CompletedDevField` and `CompletedGatewayField`. Can we do that instead of that generic field? I know attempted field has that but I would prefer to not remove the validation that we had before, just adjust it.

Yeah I can do that!

nm3224:
@vishpillai123 @bdewilde tests are failing here- does anyone know how to fix this? I think the CSV it's using for unit tests isn't aligned with the new expected inputs

vishpillai123:
> @vishpillai123 @bdewilde tests are failing here- does anyone know how to fix this? I think the CSV it's using for unit tests isn't aligned with the new expected inputs

We also need to update the unit tests when you update the code, as well. Can you take a look at what we need to change here so that we can get this to pass? `test_read_raw_pdp_course_data`

nm3224:
@vishpillai123 hey Vish, this is ready to merge!

vishpillai123 (approved):
nice job debugging the unit tests!

PR #162, opened 2025-05-15
Add custom converter function

kaylawilding:
Add handling of custom converter functions for pdp schoools. 

## changes
- Check to see if the school has a custom converter function and use that. 
- Update code in pipeline to match latest package code changes
- Update references to config file
- Update datasets in config file (Vish I believe the PDPProjectConfig file didnt need updated here since the format under datasets still worked)
- Update schema to pull from catalog rather than private sst repo 
- For now data is coming from catalog instead of buckets because of permission issue, will need to update this once Emma sorts out the issue

## questions
None


pipelines/inference/pdp_inference/workflow_asset_bundle/resources/github_sourced_pdp_inference_pipeline.yml
```diff
@@ -205,26 +205,26 @@ resources:
       git_source:
         git_url: https://github.com/datakind/student-success-tool
         git_provider: gitHub
-        git_branch: develop
+        git_branch: add-custom-converter-function
```
vishpillai123, line 208: This branch will get deleted though when we merge no?

pipelines/institution_id/config-TEMPLATE.toml
```diff
@@ -11,13 +11,15 @@ pred_prob_col = "pred_prob"
 pos_label = true
 random_state = 12345
 
-[datasets.ingested_YYYYMMDD]
+[datasets.bronze]
```
vishpillai123, line 14: we do need to update the Dataset Config in the PDPProject Config right?

If this works without the dataset config, then that's fine, but I would prefer that are less optional fields here and I think we need to clean up this part of the config in a later update.

```
class DatasetConfig(pyd.BaseModel):
    raw_course: "DatasetIOConfig"
    raw_cohort: "DatasetIOConfig"
    preprocessed: t.Optional["DatasetIOConfig"] = None
    predictions: t.Optional["DatasetIOConfig"] = None
    finalized: t.Optional["DatasetIOConfig"] = None
```

vishpillai123 (commented):
For the config stuff, if it works as is, that's fine since we're on a time crunch. I do think the current DatasetConfig is too flexible.

pipelines/institution_id/config-TEMPLATE.toml
vishpillai123: Do we need to add advisor_output under here?

PR #164, opened 2025-05-17
making features more human readable and clarifying details

nm3224:
## changes & context
- changed "avg" to average and "std. dev." to standard deviation and "min." to minimum to have outputs be more human-readable and clear to advisors
- clarified difference between cc courses and cd courses which wasn't there before
- assuming you can only take a course once in a given term, clarifying this in the human readable name


src/student_success_tool/assets/pdp/features_table.toml
```diff
@@ -8,20 +8,20 @@ num_credits_earned = { name = "number of credits earned this term" }
...
+course_level_std = { name = "standard deviation in course levels this term" }
+course_grade_numeric_mean = { name = "average numeric course grade this term" }
+course_grade_numeric_std = { name = "standard deviation in numeric course grades this term" }
+section_num_students_enrolled_mean = { name = "average number of students enrolled per section this term" }
+section_num_students_enrolled_std = { name = "standard deviation in number of students enrolled per section this term" }
 num_courses_core_course_y = { name = "number of core courses taken this term" }
 num_courses_enrolled_at_other_institution_s_y = { name = "number of courses taken this term while enrolled at another institution" }
 'num_courses_course_subject_area_(\d+)' = { name = "number of courses taken in subject area {} this term" }
-'num_courses_course_id_(.*)' = { name = "number of times course '{}' taken this term" }
+'num_courses_course_id_(.*)' = { name = "course '{}' taken this term" }
```
bdewilde, line 20: technically it is in fact the number of times the course is taken. is every value == 1?

src/student_success_tool/assets/pdp/features_table.toml
```diff
@@ -8,20 +8,20 @@ num_credits_earned = { name = "number of credits earned this term" }
...
-course_grade_numeric_mean = { name = "avg. numeric course grade this term" }
-course_grade_numeric_std = { name = "std. dev. in numeric course grades this term" }
-section_num_students_enrolled_mean = { name = "avg. number of students enrolled per section this term" }
-section_num_students_enrolled_std = { name = "std. dev. in number of students enrolled per section this term" }
+course_level_mean = { name = "average course level this term" }
+course_level_std = { name = "standard deviation in course levels this term" }
+course_grade_numeric_mean = { name = "average numeric course grade this term" }
+course_grade_numeric_std = { name = "standard deviation in numeric course grades this term" }
+section_num_students_enrolled_mean = { name = "average number of students enrolled per section this term" }
+section_num_students_enrolled_std = { name = "standard deviation in number of students enrolled per section this term" }
```
bdewilde, line 16: these are widely used abbreviations for common statistical terms. _is_ this more human readable, or just more verbose?

src/student_success_tool/assets/pdp/features_table.toml
nm3224: in the vast majority of cases, the "per term" count for a course will likely either be 1 or 0. This feature specifically confused Rutgers upon presentation b/c they didn't understand how u can take a course multiple times in 1 term. Generally I think this language is okay-- for minority schools where it's possible to take it twice, we can address with them directly, but keep it simpler for the broader majority

src/student_success_tool/assets/pdp/features_table.toml
nm3224: I think it's better to just be explicit. It might seem silly but many people have little to no statistical background and it's better to assume they know nothing than make any assumptions about anything technical

kaylawilding: Yes I agree with spelling these out

src/student_success_tool/assets/pdp/features_table.toml
bdewilde: agreed that this is potentially confusing, it's one of the reasons i advocated for turning this set of features into boolean "has-taken" equivalents. i don't know if/how that ended up happening in a PR that kayla submitted a while back ...

kaylawilding (approved):

bdewilde (commented):
Changes in `features_table.toml` seem fine, but what's `features_table_updated.toml`?

bdewilde:
> > > Changes in `features_table.toml` seem fine, but what's `features_table_updated.toml`?
> > 
> > 
> > I kept the old one just incase
> 
> @nm3224 nice job on the updated table and adding the longer descriptions. I do think it will get confusing which table to use, and I would prefer if we removed the original features table. With version history, we can always go back if we see any issues with the table you created. Can we create another PR to rename `features_table_updated.toml` to `features_table.toml` and remove the original `features_table.toml` before our Version 3.0 release this morning?

+1 to all this! Thanks for flagging @vishpillai123

PR #165, opened 2025-05-19
adding manual dependencies to train/eval notebooks & remove pin to develop

vishpillai123:
Training, eval, and register model card notebooks do not work without installing manual dependencies due to conflicts that AutoML has with our SST package. The temporary fix is to add these dependencies to the top of each of these notebooks so we can properly load and run our models trained in AutoML.  

I also chose to update the language around how "it's okay to pin to develop or a feature branch". We want to move forward to only pinning to versions so that our pipelines do not break. I also chose to remove the pip install of develop, so that we can move towards pinning to versions for schools. 

## changes
- Added overrides of pandas 1.5.3 and scikit-learn 1.3.0 to training, evaluation, and registration (+model cards) notebooks. These have worked for me whenever I've used them for testing and for my schools.
- Added language clarifying why we're doing manual dependencies
- Removed language around how it's ok to pin to develop.
- Removed `# %pip install git+https://github.com/datakind/student-success-tool.git@develop` since we want to curb that behavior. Using develop directly will lead to pipelines breaking. Of course, if you want to test our a branch, you can but I'd rather not have it in the notebooks.

## context
- Training, eval, and register notebooks (templates no.2 through 4) **do not** work without manual dependency overrides due to AutoML being in conflict with our SST package. Even installing our package with `--no-deps` (previous recommendation) has not worked for me.

## questions
Any issues with this?


bdewilde:
@vishpillai123 Can you be more specific about "Even installing our package with --no-deps (previous recommendation) has not worked for me." ?

vishpillai123:
> @vishpillai123 Can you be more specific about "Even installing our package with --no-deps (previous recommendation) has not worked for me." ?

Yes, so if I try to run the following, with no manual overrides:
`%pip install "student-success-tool==0.3.1" --no-deps`

I then, try to import our packages and I get the following error:
`from student_success_tool import configs, dataio, modeling, utils`
Error:
```
AttributeError: module 'pydantic' has no attribute 'computed_field'
File <command-7351897682668367>, line 9
      6 from databricks.connect import DatabricksSession
      7 from databricks.sdk.runtime import dbutils
----> 9 from student_success_tool import configs, dataio, modeling, utils
```

This is only avoided by installing manual dependencies on top of our SST package.

bdewilde:
Got it. Just confirming that [this](https://github.com/datakind/student-success-tool/blob/develop/src/student_success_tool/utils/databricks.py#L28-L84) doesn't solve the issue?

PR #177, opened 2025-05-20
Update datasets classes

kaylawilding:
Update the config class to accept the new dataset structure of bronze, silver, and gold tables


vishpillai123 (approved):
perfect thanks for doing this!

bdewilde:
Hey @vishpillai123 , thanks for explaining! I think I gently disagree with some of the decisions here, but they make much more sense now that I understand the context in which they'll be used. (This PR has no description / details ... so I had no way to know!)

PR #179, opened 2025-05-20
refactoring PDP templates to v0.3.1

vishpillai123:
Updated templates to most recent version v0.3.1. Per our recent team alignment, we will be using these notebooks only interactively since PDP inference will be through shared pipeline in web app.

## changes
- Updated to v0.3.1.
- Removed outdated comments and code.
- Removed language around "it is okay to add to config later and hardcode now". 
   - When users submit these notebooks for PR, we want to ensure all information is set in the config. 
   - All school-specific info needs to be in the config. No parameters should be specified from the notebooks.
- Added saving of preprocessed, modeling, and advisor output datasets.
- Removed saving of validated datasets in EDA notebook since these aren't being used anywhere else. Validated datasets are created using our converter functions in the preprocess-data notebook. 
   - We also want all datasets that are being saved to be standardized across all PDP schools, which is why we updated the config to reflect that.

## context
- Templates need to be updated to latest version and our most recent process.

## questions
- v0.3.1 looks solid, only thing that needs to be patched up is our PDP features table. We are missing several features. How do we identify all the features that are missing? One of my schools was missing the following: `took_course_course_level_0_1`, `term_program_of_study_area`, `term_program_of_study`, `took_course_course_type_cc_cd`. This needs to be patched up before our June release.

### DIFF
--- a/pipelines/institution_id/00-data-assessment-TEMPLATE.py
+++ b/pipelines/institution_id/00-data-assessment-TEMPLATE.py
@@ -25,7 +25,7 @@
 
 # install dependencies, of which most/all should come through our 1st-party SST package
 
-# %pip install "student-success-tool == 0.2.0"
+# %pip install "student-success-tool == 0.3.0"
 
 # COMMAND ----------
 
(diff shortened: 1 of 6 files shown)


kaylawilding (approved):
Changes look good to templates. I agree the features table is outstanding on 0.3.1. When I do that, I will update the templates to pull 0.3.2. I was hitting this error all of a sudden on the config template validation, did you run into this? It seems similar to the different data classes when we did bronze, silver, gold that it thinks each combination is required. 
Screen Shot 2025-05-21 at 8 30 49 PM

PR #183, opened 2025-05-21
feat: added tables for support distribution, top shap values, feature…

Mesh-ach:
… importance

**Added support score distribution and top feature importance tables**

## changes

-  Implemented support_score_distribution_table() to generate binned support score histograms
-  logic to bin scores starting from 0.1 to 1.0 in 0.1 increments

## context
Changes to generate explainability outputs for the front end:

-  support_score_distribution outputs the support score distribution table
-  top_10_features and inference_shap_feature_importance support top feature explanations

## questions
No question at this time

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1210345786296114
  - https://app.asana.com/0/0/1210321908468809


src/student_success_tool/modeling/inference.py
```diff
@@ -293,3 +293,100 @@ def calculate_shap_values(
...
+            unique_ids,
+            pred_probs,
+            shap_values.values,
+            n_features=inference_params["num_top_features"],
+            features_table=features_table,
+            needs_support_threshold_prob=inference_params["min_prob_pos_label"],
+        )
+
+        # --- Bin support scores for histogram (e.g., 0.0 to 1.0 in 0.1 steps) ---
+        bins = np.arange(0.1, 1.1, 0.1)
```
vishpillai123, line 356: We may want to have a smaller bin size per the Figma design, maybe 0.05? I think they may even did 0.02. Anyways, we can change this later too.

src/student_success_tool/modeling/inference.py
```diff
@@ -293,3 +293,100 @@ def calculate_shap_values(
...
+    unique_ids,
+    pred_probs,
+    shap_values,
+    inference_params,
+    features_table,
+    model_feature_names,
+):
+    """
+    Selects top features to display and store
+    """
```
vishpillai123, line 342: The description here I think is maybe misleading? I know that this table needs the support scores, but I don't think we need the features?

src/student_success_tool/modeling/inference.py
```diff
@@ -293,3 +293,100 @@ def calculate_shap_values(
         # reattach student ids to their shap values
         .assign(**{student_id_col: student_ids})
     )
+
+
+def top_shap_features(
```
vishpillai123, line 298: From what I understand, this table is mostly used just to just pull the top features from the summary shap plot. Just checking that we don't need `select_top_features_for_display` for this one right?

src/student_success_tool/modeling/inference.py
```diff
@@ -293,3 +293,100 @@ def calculate_shap_values(
...
+        .melt(id_vars="student_id", var_name="feature_name", value_name="shap_value")
+    )
+
+    feature_long = features.assign(student_id=unique_ids.values).melt(
+        id_vars="student_id", var_name="feature_name", value_name="feature_value"
+    )
+
+    summary_df = shap_long.merge(feature_long, on=["student_id", "feature_name"])
+
+    top_10 = (
```
vishpillai123, line 318: Perhaps we can make the top X features a parameter here? Not sure if we will later get the request from leadership to show top 15 vs. top 10 etc. We can default this to 10.

vishpillai123 (approved):
Nice job!! This is production ready and I like how you used inference code within the pdp_model_inference pipeline. 

Just added some comments/feedback, but overall I think we're good to move forward.

Mesh-ach:
Hey Vish, thanks for the review. I’ve seen the comments, I’m working on creating sample tables first. I’ll respond to the comments soon and send the final PR later in the day.

src/student_success_tool/modeling/inference.py
Mesh-ach: Yeah, we can always readjust this if they don't like it

src/student_success_tool/modeling/inference.py
Mesh-ach: Yeah I agree, adjusting this in next PR

src/student_success_tool/modeling/registration.py
```diff
@@ -54,6 +60,202 @@ def register_mlflow_model(
     LOGGER.info("model version successfully registered at '%s'", model_path)
 
 
+def log_confusion_matrix(
```
vishpillai123, line 63: great job with this and roc below, though can we move them to either `inference` or `evaluation`?

src/student_success_tool/modeling/registration.py
```diff
@@ -54,6 +60,202 @@ def register_mlflow_model(
...
+        )
+
+    except mlflow.exceptions.MlflowException as e:
+        raise RuntimeError(f"MLflow error while retrieving run {run_id}: {e}")
+    except Exception:
+        LOGGER.exception("Failed to compute or store confusion matrix.")
+        raise
+
+
+def log_roc_table(
```
vishpillai123, line 150: you did this fast! haha nice job!!

src/student_success_tool/modeling/registration.py
```diff
@@ -54,6 +60,202 @@ def register_mlflow_model(
...
+        data_run_id = runs_df[runs_df["tags.mlflow.runName"] == data_run_tag][
+            "run_id"
+        ].item()
+
+        # Load test data
+        artifact_path = mlflow.artifacts.download_artifacts(
+            run_id=data_run_id, artifact_path="data", dst_path=tmp_dir
+        )
+        df = pd.read_parquet(os.path.join(artifact_path, "training_data"))
+        test_df = df[df["_automl_split_col_0000"] == "test"].copy()
```
vishpillai123, line 202: We can retrieve this from the config: cfg.split_col, typically. If it doesn't exist then I believe it's the automl split that you specified here.

Burton wrote the following code in the templates that I think you could also use too (this is in `02-train-model-TEMPLATE.py`)
```
# NOTE: AutoML generates a split column if not manually specified
split_col = training_params.get("split_col", "_automl_split_col_0000")
```

src/student_success_tool/modeling/registration.py
```diff
@@ -54,6 +60,202 @@ def register_mlflow_model(
...
+
+        spark = (
+            SparkSession.builder.master("local[*]").appName("Fallback").getOrCreate()
+        )
+
+    data_run_tag = "Training Data Storage and Analysis"
+    table_path = f"{catalog}.{institution_id}_gold.sample_training_{run_id}_roc_curve"
+    tmp_dir = f"/tmp/{uuid.uuid4()}"  # unique tmp path
+
+    try:
```
vishpillai123, line 182: This will be run in the context of a run_id in the evaluation code. So, I don't think we need this client code with searching for the run from the experiment. You can see how we do this for our other code in `evaluation` and `bias_detection`. We can do the following:

```
with mlflow.start_run(run_id=cfg.model.run_id): 
     log_confusion_matrix(...)
```


I do think we could assert that there is an active run. I usually assume that a run is active, but I think a better practice is to add an assertion, like below.
`assert mlflow.active_run(), "No active MLflow run. Please call this within an MLflow run context."
`

Or if you prefer to have your implementation, then that's fair as well! Curious about your thoughts too on best practice.

src/student_success_tool/modeling/registration.py
```diff
@@ -54,6 +60,202 @@ def register_mlflow_model(
...
+        test_df = df[df["_automl_split_col_0000"] == "test"].copy()
+
+        # Load model + features
+        model = mlflow.sklearn.load_model(f"runs:/{run_id}/model")
+        feature_names: List[str] = model.named_steps["column_selector"].get_params()[
+            "cols"
+        ]
+
+        # Infer target column
+        excluded_cols = set(feature_names + ["_automl_split_col_0000"])
```
vishpillai123, line 211: We don't need to infer the target column actually! We can retrieve it from the config: `cfg.target_col`, typically.

pipelines/tasks/pdp_model_inference/pdp_model_inference.py
```diff
@@ -259,6 +259,78 @@ def calculate_shap_values(
             logging.error("Error during SHAP value calculation: %s", e)
             raise
 
+    def top_n_features(
```
vishpillai123, line 262: Maybe we can `n` here as well, and set it to 10? And then we can feed that into `inference.top_shap_features`.

pipelines/tasks/pdp_model_inference/pdp_model_inference.py
```diff
@@ -259,6 +259,78 @@ def calculate_shap_values(
...
+        shap_values: npt.NDArray[np.float64],
+    ) -> pd.DataFrame:
+        try:
+            top_n_shap_features = inference.top_shap_features(
+                features, unique_ids, shap_values
+            )
+            return top_n_shap_features
+
+        except Exception as e:
+            logging.error("Error computing top 10 shap features table: %s", e)
```
vishpillai123, line 274:
```suggestion
            logging.error("Error computing top %d shap features table: %s", n, e)
```

vishpillai123 (commented):
A few suggestions, but we're close. let me know what you think!

src/student_success_tool/modeling/registration.py
Mesh-ach: Moved it to evaluation

src/student_success_tool/modeling/registration.py
Mesh-ach: Turned it into an argument instead since other functions also accepts split_col as an argument

PR #185, opened 2025-05-24
upgrade `pandera` and consolidate parsing logic

bdewilde:
## changes

- upgrades `pandera` from v0.23 => v0.24, in order to pull in a couple key bugfixes
- updates to `pandera.pandas` imports, as now recommended in v0.24
- moves and consolidates raw data parsing logic in the data schemas themselves, rather than splitting across the read func and the schema, as enabled by v0.24

## context

The awkward data parser split between read-func and schema was known and blocked by a couple bugs in v0.23; upgrading to v0.24 allowed us to resolve this issue. There are a few nice benefits:

- logic in the dataio read funcs is minimized, so the raw data is "more raw"; this better separates functionality, and makes it easier to write appropriate converter funcs (no need to account for hidden parsing logic)
- parsing logic in the schemas is consolidated and better separated as well: the schemas are _the_ place where raw data transformations occur, and column-level parsers for setting categoricals are now, actually, only for setting categories rather than also parsing the input strings

Overall this isn't a _huge_ deal, but it was a notable, largely hidden spot in the code that could cause confusion and maintenance headaches.

## questions

- does this break anything that wouldn't be caught by unit tests? possibly related to converters or school-specific schema overrides....


src/student_success_tool/dataio/schemas/pdp/raw_cohort.py
```diff
@@ -252,67 +252,103 @@ def rename_student_id_col(cls, df):
...
-    def strip_and_uppercase_strings(cls, series):
-        return series.str.strip().str.upper()
+    @pda.dataframe_parser
+    def strip_upper_string_values(cls, df):
+        """Strip/uuppercase string values for some cols to avoid inconsistency later on."""
+        return df.assign(
+            **{
+                col: ft.partial(_strip_upper_string_values, col=col)
+                for col in (
+                    "cohort_term",
```
vishpillai123, line 262: why are we doing strip/uppercase on more columns now?

vishpillai123 (commented):
Reading of raw data is a lot more cleaned up as you have said, while there are more transformations happening at schema level. 

I do think that this might be more difficult to maintain in theory though. If there are updates to schemas, then we would need to update these transformations and add/remove columns right? Our previous code with pandera 0.23 appears to apply many of these transformations at the initial reading of raw data, which I think is more maintainable especially since these raw tables all share these transformations regardless of schema-level differences.

As far as your question on whether this breaks anything outside of the unit tests, I would prefer that you test that out in this branch before we merge since this is a dependency upgrade and _not_ critical to either our onboarding notebooks or the inference pipeline.

src/student_success_tool/dataio/schemas/pdp/raw_cohort.py
bdewilde: For the same reason we're applying these transformations to some columns, it makes sense to apply them to other columns: consistency of outputs, robustness to raw data weirdness, better guarantees about categoricals, ...

bdewilde:
> I do think that this might be more difficult to maintain in theory though. If there are updates to schemas, then we would need to update these transformations and add/remove columns right? Our previous code with pandera 0.23 appears to apply many of these transformations at the initial reading of raw data, which I think is more maintainable especially since these raw tables all share these transformations regardless of schema-level differences.

I strongly disagree with this view. These changes bring *all of the raw data parsing into the schemas themselves, rather than splitting it across data i/o and the schemas. By co-locating parsing logic alongside validation logic, it's easier and more obvious to make changes to them in unison. (That's why `pandera` is designed to do both! Practically speaking, it doesn't make sense to separate the two tasks.)

PR #192, opened 2025-05-29
fix: adjusting non_feature_cols in config and templates

vishpillai123:
Currently, `non_feature_cols` isn't included in model_dump(), so unless it's specified in the config (which it shouldn't because it's a computed column), then it won't be included when performing feature selection. This column should not be both a computed field and an optional field. It is technically not optional, as feature selection doesn't work properly without it (student group cols or split/sample_weight gets dropped, etc.)

## changes
- Removed `non_feature_cols` as an optional field in FeatureSelectionConfig. It is already included as a computed field in the parent model. 
- We can not add the computed field under the FeatureSelectionConfig since it makes calls to parameters that sit in the parent model (`student_id`, `sample_weight_col`, etc.). Technically, `non_feature_cols` is also applicable to evaluation as well, so we can just leave the computed field as is.
- In our templates, we can set `non_feature_cols` directly after model_dump() in selection_params since it's needed for feature selection. 

## context
- The column `non_feature_cols` isn't included in model_dump(), but it's needed for feature selection. 
- We're aiming for DRY with our configs, so this change will help with that.
- Setting `non_feature_cols` directly in our templates is cleanest way I thought of doing this. Long-term we can probably find a way to include this in directly from model_dump(), but it requires more manipulation & work with our config.


nm3224 (commented):
Is there a way to get it to not have "None" under the selection params for non_feature_cols?

nm3224:
ahh got it okay, that makes sense!

PR #196, opened 2025-06-04
Remove first last terms

kaylawilding:
Remove the nth_student_term wrapper functions so we only have the nth student term fn, but can now specify type to subset the data. 

## changes
- Removed the wrappers in the checkpoint functions
- Added a helper to subset based on the type now specified in the config
- Updated config and config validation
- Updated the preprocessing template so no longer necessary to select a checkpoint fn (there is only one now), but you do need to add the appropriate parameters based on type (this is not new though)

## context
There were several wrapper functions for the checkpoint function of nth_student_terms that specified for example that first or last term and how to subset the data. 

Instead, we are just specifying the options for first and last. Also, just using a helper fn to subset the data. This should help with extensibility and maintenance of the code. Additionally, I think this will help us in converting to a py file as it moves the type logic inside the fn instead of having to put it in the pipeline of which fn to use.

## questions
@vishpillai123, can you check youre happy with edits to model cards?


src/student_success_tool/preprocessing/checkpoints/pdp.py
```diff
@@ -11,15 +11,21 @@
...
+    type: t.Literal[
+        "all", "num_credits_earned", "within_cohort", "enrollment_year"
+    ] = "all",
+    enrollment_year: t.Optional[int] = None,
     student_id_cols: str | list[str] = "student_id",
     sort_cols: str | list[str] = "term_rank",
     include_cols: t.Optional[list[str]] = None,
     term_is_pre_cohort_col: str = "term_is_pre_cohort",
-    exclude_pre_cohort_terms: bool = True,
+    exclude_pre_cohort_terms: bool = False,
```
vishpillai123, line 23: I thought we wanted the excludes to default to True?

src/student_success_tool/preprocessing/checkpoints/pdp.py
```diff
@@ -11,15 +11,21 @@
 def nth_student_terms(
```
vishpillai123, line 11: This function has become a lot more concise and efficient code-wise, but it seems more complicated to use. Docstring could be helpful to provide some more clarity.

src/student_success_tool/reporting/sections/pdp/attribute_sections.py
```diff
@@ -100,19 +100,16 @@ def checkpoint():
         """
         checkpoint_type = card.cfg.preprocessing.checkpoint.type_
         base_message = "The model makes this prediction when the student has"
-        if checkpoint_type == "nth":
-            n_ckpt = card.cfg.preprocessing.checkpoint.n
+        n_ckpt = card.cfg.preprocessing.checkpoint.n 
+
+        if checkpoint_type == "all":
```
vishpillai123, line 105: I created a PR yesterday, maybe we can include this? If you want to revert these changes, and I can also add on to what I have in this PR. [#195]

PR #197, opened 2025-06-04
Remove na target values

kaylawilding:
Remove NAs from the dataset if the target is incalculable.

## changes
Updated the compute target function for grad, retention, and credits earned to drop the NA target obs before returning. Updated tests to not expect NA values in the output.

## context
When we compute the targets, previously, students with an NA value due to the fact they did not meet the eligibility criteria or have enough time left were included. Now, the code will drop these students that do not have a calculable target from the dataset. 
The NAs in the datasets were breaking graphing among other functions with this dataset.

## questions
none

### DIFF
--- a/mlruns/0/meta.yaml
+++ b/mlruns/0/meta.yaml
@@ -1,6 +1,0 @@
-artifact_location: file:///Users/kaylawilding/Documents/Apps/Portfolio/Github/student-success-tool/mlruns/0
-creation_time: 1749059674522
-experiment_id: '0'
-last_update_time: 1749059674522
-lifecycle_stage: active
-name: Default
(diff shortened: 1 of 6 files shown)


bdewilde:
@kaylawilding I spent a lot of time and care making the targets correctly assign `NA` to unlabel-able students, and I'd be quite sad to see that ripped out like this. There are use cases where you want to be able to see / analyze students with and without targets. Would it be possible just to filter rows outside of these functions to non-null only, as the use case demands?

vishpillai123:
> @kaylawilding I spent a lot of time and care making the targets correctly assign `NA` to unlabel-able students, and I'd be quite sad to see that ripped out like this. There are use cases where you want to be able to see / analyze students with and without targets. Would it be possible just to filter rows outside of these functions to non-null only, as the use case demands?

besides doing EDA around these NA students, I don't believe isn't any use case for why we need these students. During onboarding, we filter them in the notebook. For inference, we're not computing a target directly... just running the model with the features/data given to produce a prediction.

bdewilde:
Okay, that's your call. I'm going to stop hopping in with questions and comments unless specifically asked.

vishpillai123 (approved):

PR #206, opened 2025-06-06
feat: handle mlflow exceptions better in model cards

vishpillai123:
## changes
- This PR is meant to add more robustness and better error handling in our model card module.
- When mlflow has any errors or raising of exceptions, this automatically breaks the `build` step and I want to make this more robust.
- Added note in case model version isn't found (this happens a lot because databricks client is annoying). In this case we just won't show the model version.
- When counting number of runs in an experiment, this can sometimes create a timeout issue, so we want to try to handle these cases.
- Added a test to check whether model card context gets updated correctly. This should ensure when we make updates to MC that it will build properly in our markdown template. It basically looks at anything that needs to be specified in our template and makes sure that our model cards cover everything. 

## context
- We are creating more model cards for schools and we want to have better exception handling and testing, in general.
- A school's model card errored upon build due to mlflow api issues. We want to ensure that this is handled properly.

## questions
- No questions


src/student_success_tool/reporting/sections/attribute_sections.py
```diff
@@ -10,6 +11,19 @@ def register_attribute_sections(card, registry):
...
+    def development_note():
+        """
+        Produce a note describing when the model was developed and listing the
+        model version (if available).
+        """
+        version_number = card.context.get("version_number", None)
+        current_year = str(datetime.now().year)
+        if version_number:
+            return f"Developed by DataKind in {current_year}, Model Version {version_number}"
+        else:
```
kaylawilding, line 24: Curious why you decided on year for when the model was developed? Should we be more precise?

src/student_success_tool/reporting/utils/utils.py
```diff
@@ -140,5 +158,58 @@ def list_paths_in_directory(run_id: str, directory: str) -> t.List[str]:
...
+        )
+        return [artifact.path for artifact in artifacts]
+    except Exception as e:
+        LOGGER.error(
+            f"Failed to list artifacts in directory '{directory}' for run '{run_id}': {e}"
+        )
+        return []
+
+
+def safe_count_runs(
```
kaylawilding, line 173: Weird you would think MLflow would track this without having to do this pagination but cool work around. Silly MLflow.

kaylawilding (approved):
Looks to me like you did a nice job adding error handling and updating the runs numbers. I don't know though what the issue was that prompted this PR, so @nm3224 should probably review.

src/student_success_tool/reporting/utils/utils.py
vishpillai123: Yeah it's dumb. This is what GPT recommended since I thought we had a timeout issue but it looks like it was just the GCP vs. Azure stuff. Anyways, this safe count should still be good in case there are weird errors with mlflow API.

vishpillai123:
> Looks to me like you did a nice job adding error handling and updating the runs numbers. I don't know though what the issue was that prompted this PR, so @nm3224 should probably review.

I think it was miscommunication since it looks like the error was seen on GCP. Either way, I tried to cast a wide net with the error handling here with mlflow that should be generally helpful for us.

PR #211, opened 2025-06-10
Updating feature table

nm3224:
## changes & context
 updating feature table names


rachelauryn:
Looping in @tarachiatovich to a couple PRs so she can get some visibility into our current work!

nm3224:
> @nm3224 hi, is this ready for review?
> 
> Also could you provide some more background on this? It appears that we're just missing these features that you're adding based on the diff, and is that it?

I think @kaylawilding was going to fix the naming issues with mapping those custom course “within 12 credits” and “cummax” features on this branch

PR #217, opened 2025-06-20
Add features table to top shap features table

kaylawilding:
Add a parameter to use the features tabl in the top shap features function

## changes
Add a parameter to use the features tabl in the top shap features function

## context
We want to be able to output the human readable names in the inference tables of the top shap features, and in order to do that we needed to add a way to use the features table in the top shap features function in inference

## questions
None


src/student_success_tool/modeling/inference.py
```diff
@@ -383,6 +385,11 @@ def top_shap_features(
 
     top_features = summary_df[summary_df["feature_name"].isin(top_n_features)].copy()
 
+    if features_table is not None:
+        top_features["feature_name"] = top_features["feature_name"].apply(
+            lambda feature: _get_mapped_feature_name(feature, features_table)
+        )
+
```
Mesh-ach, line 392:
```suggestion
top_features['feature_value'] = top_features['feature_value'].astype(str)

```

Mesh-ach (approved):

PR #223, opened 2025-06-25
feature table was missing some features, I added them in

nm3224:
## changes & context

(NM) List of features I added to features_table toml:

-  'num_courses_course_grade_audit',
-  'num_courses_course_grade_i',
-  'num_courses_course_grade_other',
-  'num_courses_course_grade_w',
-  'frac_courses_course_grade_audit',
-  'frac_courses_course_grade_i',
-  'frac_courses_course_grade_other',
-  'frac_courses_course_grade_w',
-  'num_courses_course_grade_audit_cumfrac',
-  'num_courses_course_grade_i_cumfrac',
-  'num_courses_course_grade_other_cumfrac',
-  'num_courses_course_grade_w_cumfrac',

(KW) Feature engineering changes: 

1. Add replacing students within_12_credit features as missing if their cum num credits < 12 and have not yet taken the course into the processing.pdp.clean_up_labeled_dataset_cols_and_vals feature
2. Add boolean type features to be included in VIF code for selection in this line df_features = df.select_dtypes(include=“number”) in drop_collinear_features_iteratively function in feature_selection
3. Edit config to include list of lists for subject areas/key course ids
4. Limit bool features to only subject area/key course ids and remove applicable features from feature toml

(KW) Questions:
Are you happy with this logic tree for #1 ?
"To prevent data leakage, students that have not reached the 12 credits and not taken the course  by the checkpoint term (which this data is limited to at the time of this function), will have the applicable in_12_credits columns set to null. Students that took the course will remain True, regardless of credits achieved."

### DIFF
--- a/src/student_success_tool/preprocessing/pdp.py
+++ b/src/student_success_tool/preprocessing/pdp.py
@@ -253,12 +253,10 @@ def clean_up_labeled_dataset_cols_and_vals(
     credit_pattern = re.compile(rf"in_{num_credit_check}_credits")
     # To prevent data leakage, students that have not reached the 12 credits and not taken the course
     # by the checkpoint term (which this data is limited to at the time of this function),
-    # will have the applicable in_12_credits columns set to null. 
+    # will have the applicable in_12_credits columns set to null.
     for col in df.columns:
         if credit_pattern.search(col):
-            df[col] = df[col].mask(
-                (df[num_credits_col] < num_credit_check)
-            )
+            df[col] = df[col].mask((df[num_credits_col] < num_credit_check))
 
     return (
         # drop many columns that *should not* be included as features in a model


nm3224:
> @nm3224 I cant add you as a reviewer since this was originally your PR, but could you also take a look?

@kaylawilding yes I’ll take a look Monday!

src/student_success_tool/preprocessing/pdp.py
```diff
@@ -238,11 +238,29 @@ def add_empty_cols_if_missing(
...
     any values corresponding to time after a student's current year of enrollment.
+
+    Args:
+        df: DataFrame as created with features and targets and limited to the checkpoint term.
+        num_credits_col: Name of the column containing cumulative earned credits.
     """
+    num_credit_check = constants.DEFAULT_COURSE_CREDIT_CHECK
+    credit_pattern = re.compile(rf"in_{num_credit_check}_credits")
+    # To prevent data leakage, students that have not reached the 12 credits and not taken the course
+    # by the checkpoint term (which this data is limited to at the time of this function),
```
nm3224, line 255: @kaylawilding why are we only setting to null if the column is False? Even if the value is True, it should still be null if the student hasn't yet reached 12 credits, b/c that's the data leakage then... am I misunderstanding?

nm3224: if you haven't yet reached 12 credits, but you took the course after the checkpoint - you would still have a true value, right? and that is info beyond the checkpoint which is leakage

nm3224:
> > One more thing to check: for variables like
> > 
> > * Years to latest associate’s degree at cohort inst.
> > * Years to latest certificate at cohort inst.
> > * First year to certificate at other inst.
> > * First year to associate’s degree or certificate at other inst.
> > * First year to associate’s degree at other inst.
> > 
> > Do we have logic in place to only include degrees or certificates that were achieved **_before_** the checkpoint?
> 
> We drop the following
> 
> * ```
>   "years_to_bachelors_at_cohort_inst",
>   ```
> * ```
>   "years_to_bachelor_at_other_inst",
>   ```
> * ```
>   "years_to_associates_or_certificate_at_cohort_inst",
>   ```
> * ```
>   "years_to_associates_or_certificate_at_other_inst",
>   ```
> * ```
>   "years_of_last_enrollment_at_cohort_institution",
>   ```
> * ```
>   "years_of_last_enrollment_at_other_institution",
>   ```
> 
> and we keep "first year to X credential at Y inst" values if they occurred in any year prior to the current year of enrollment; otherwise, set to null in this case, the values themselves represent years.

@kaylawilding we should note tho that PDP data has changed since this was written and the columns are now different so we may need to edit this

kaylawilding (approved):

PR #228, opened 2025-06-30
feat/adding in custom school templates

vishpillai123:
## changes
- Added new templates for custom schools.
- Also for now added the HITL notebook that I've been using for custom schools. It goes over all the checks that I've been doing such as distribution, sanity checks for certain students, etc. We can also add more to this too. But not sure how this is going to change once we integrate with the web app. We will change accordingly once we get there.
- This is just a skeleton and will likely be different for schools depending on the number of models they have.

## context
- Custom schools really need templates.


vishpillai123:
adding @kaylawilding as optional reviewer

PR #234, opened 2025-07-01
fixing table

nm3224:
## changes & context
- some schools have "_" in their prefixes
- order has to be changed for cumfrac and non cumfrac otherwise it just uses the first one it sees


src/student_success_tool/assets/pdp/features_table.toml
```diff
@@ -137,8 +137,8 @@ num_courses_course_grade_w = { name = "number of W (withdrawn) grades earned thi
...
+'num_courses_course_id_(.*)' = { name = "number of times course '{}' taken this term", desc = "This feature represents if a student has taken in a specific course in a given (usually the checkpoint) term, based on the course ID. A value of 1 for this feature would mean that a student took the specified course in a given (usually the checkpoint) term, while a value of 0 for this feature would mean that the student did not take the specified course in a given (usually the checkpoint) term. In some modular or non-standard academic systems, a student might take the same course multiple times within a single term, leading this value to be greater than 1. This field is usually custom-engineered in relation to key courses known to be predictive of students' academic performances at the specific institution." }
```
kaylawilding, line 141: You should add $ at the end of this one so 
'num_courses_course_id_(.*)$'
But actually lets just add it to all of them, please. So any regex names that dont have $ at the end can you add?

src/student_success_tool/assets/pdp/features_table.toml
```diff
@@ -279,9 +279,9 @@ term_is_while_student_enrolled_at_other_inst_cumsum = { name = "number of terms
...
-'^took_course_id_([a-z]+)_(\d+)_cummax_in_12_creds$' = { name = "course with id '{} {}' taken within student’s first 12 credits", desc = "This feature represents if a student has taken in a specific course in their first 12 credits, based on the course ID. A value of 1 for this feature would mean that a student took the specified course in their first 12 credits, while a value of 0 for this feature would mean that the student did not take the specified course in their first 12 credits." }
+'^took_course_id_([a-z_]+)_(\d+)$' = { name = "course with id '{}{}' taken this term", desc = "This feature represents if a student has taken in a specific course in a given (usually the checkpoint) term, based on the course ID. A value of 1 for this feature would mean that a student took the specified course in a given (usually the checkpoint) term, while a value of 0 for this feature would mean that the student did not take the specified course in a given (usually the checkpoint) term." }
```
kaylawilding: actually do you want to try changing these to '^took_course_id_(.*)$', '^took_course_id_(.*)_cummax$', '^took_course_id_(.*)_cummax_in_12_creds$'? that will be more flexible for us. Then also in the names we will need to change it to "course with is '{}' taken so far" for these.

kaylawilding (changes requested):

vishpillai123 (approved):

PR #236, opened 2025-07-02
template update adding percents

nm3224:
## changes &  context
adding percents (%) labels on top of percent credits earned each year of enrollment graph


vishpillai123 (approved):
Hey Noreen this looks good once you

PR #239, opened 2025-07-03
fixing typo; years_to_bacheloR not years_to_bachelorS - only for "oth…

nm3224:
## changes & context

weird inconsistency i discovered in column names for first_year_to_bachelorS_at_cohort_inst and years_to_bachelorS_at_cohort_inst vs. first_year_to_bacheloR_at_other_inst and years_to_bacheloR_at_other_inst


vishpillai123 (commented):
do we know if this will be the case for all institutions? can we confirm?

vishpillai123 (approved):

PR #240, opened 2025-07-03
removing prior degree feature

nm3224:
## changes & context
fearing data leakage with this feature, deciding to remove


vishpillai123 (approved):

src/student_success_tool/preprocessing/features/pdp/student_term.py
```diff
@@ -236,25 +230,6 @@ def year_of_enrollment_at_cohort_inst(
...
 
 
-def student_has_prior_degree(
-    df: pd.DataFrame,
-    *,
-    inst: t.Literal["cohort", "other"],
-    enrollment_year_col: str = "year_of_enrollment_at_cohort_inst",
-) -> pd.Series:
-    degree_year_cols = [
-        f"first_year_to_certificate_at_{inst}_inst",
```
vishpillai123, line 246: Wait, actually for "first_year_to_certificate..", I thought we wanted to create features for these?

vishpillai123: Yeah for sure we can adjust this to have T/F for certificates. I feel like that’s what we need the most from “first year to certificate” etc. Unless we have access to the number of certificates, but otherwise a T/F feature works the best here I think.

nm3224: ok im in the process of updating it @vishpillai123

PR #242, opened 2025-07-08
Adjusted fe script

Mesh-ach:
Add support for SHAP feature metadata (short/long descriptions) using feature mapping

## changes
- Function to return short_desc and long_desc
- Updated top_shap_features and inference_shap_feature_importance to output human-readable metadata

## context
Part of a ticket https://app.asana.com/1/6325821815997/project/1205244929570055/task/1210676079287409?focus=true to resolve some endpoint bugs for Larry's frontend work

### DIFF
--- a/tests/modeling/test_inference.py
+++ b/tests/modeling/test_inference.py
@@ -417,6 +417,9 @@ def sample_data():
             "short_desc": "A short description of feature 2",
             "long_desc": "A long description of feature 2",
         },
+        "feature3": {
+            "name": "Feature 3 Name",
+        },
     }
     return features, unique_ids, shap_values, features_table
 
@@ -455,6 +458,17 @@ def test_top_shap_features_behavior(sample_data):
     print(grouped)
     assert grouped.index[0] == "Feature 1 Name"
     assert grouped.index[1] == "Feature 2 Name"
+
+    assert (
+        result["feature_short_desc"]
+        .apply(lambda x: isinstance(x, str) or x is None)
+        .all()
+    )
+    assert (
+        result["feature_long_desc"]
+        .apply(lambda x: isinstance(x, str) or x is None)
+        .all()
+    )
 
 
 def test_handles_fewer_than_10_features():
--- a/pipelines/pdp/tasks/pdp_model_inference/pdp_model_inference.py
+++ b/pipelines/pdp/tasks/pdp_model_inference/pdp_model_inference.py
@@ -335,7 +335,9 @@ class ModelInferenceTask:
                 ["readable_feature_name", "short_feature_desc", "long_feature_desc"]
             ] = shap_feature_importance["Feature Name"].apply(
                 lambda feature: pd.Series(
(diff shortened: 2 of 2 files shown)


pipelines/pdp/institution_id/03-make-predictions-TEMPLATE.py
```diff
@@ -289,6 +289,17 @@
 shap_feature_importance = inference.generate_ranked_feature_table(
```
vishpillai123, line 289: This appears to be duplicated. It looks like we are calling the `inference.generate_ranked_feature_table` above in this template?

And with that, could we add this logic below (lines 292-302) into that function?

src/student_success_tool/modeling/inference.py
```diff
@@ -261,6 +261,38 @@ def _get_mapped_feature_name(
...
+                    break
+        else:
+            feature_name = feature_col
+            short_desc = None
+            long_desc = None
+
+    return feature_name, short_desc, long_desc
+
+
 def calculate_shap_values_spark_udf(
```
vishpillai123, line 296: unfortunately, this is going to break how we use this function for custom schools (we use the same function). We don't have `short_desc` or `long_desc`, and I only use `name` for custom schools. Can we still have a fall back for that?

tests/modeling/test_inference.py
```diff
@@ -407,8 +407,16 @@ def sample_data():
         ]
     )
     features_table = {
-        "feature1": {"name": "Feature 1 Name"},
```
vishpillai123, line 410: Let's add tests using ["name"] OR ["name", "short_desc", "long_desc"].

vishpillai123:
hey @Mesh-ach, I added some feedback. Just want to give a heads up that we need these functions to also work for custom schools where we don't have "long_desc", "short_desc". Is there a way we can keep functionality for custom schools? I understand this makes it sort of complicated, but let me know what you think.

pipelines/pdp/institution_id/03-make-predictions-TEMPLATE.py
Mesh-ach: Yeah, so it's not a duplicate. The first version feeds `features_table` directly into the `inference.generate_ranked_feature_table` function, which returns the transposed feature names. What we want is to retain the original feature name and return the three other variables as well. That's why I re-initialized the function for the FE table, this way, we get the raw features back and manually feed the DataFrame into the `_get_mapped_feature_name` function.

src/student_success_tool/modeling/inference.py
Mesh-ach: Actually, I wrote a separate function for the FE script called `_get_mapped_feature_metadata` so it’s different. But after seeing your comment, I adjusted the original `_get_mapped_feature_name` function to take an optional argument that defaults to False. That way, we can keep using `_get_mapped_feature_name` as is and just pass metadata=True when we want the additional fields returned.

vishpillai123 (approved):

PR #243, opened 2025-07-09
Move cum{fn} parts of names to front

kaylawilding:
So that we can make the regex for subject area and course ids be more flexible without catching the endings, we moved the additional texxt such as the cum{fn} parts to the beginnings, so now our regex features end in the regex and can use the greediest regex

Additionally, add extra tests


uv.lock
vishpillai123, line 1: maybe it's okay to update this actually?

vishpillai123 (approved):

PR #253, opened 2025-07-15
adding custom school model cards

vishpillai123:
## changes
- Created first pass on custom model cards. 
- We are using inheritance, so a lot of the code is just reused. The design imitates PDP, but with a few additional overrides in certain sections. 
- Added test suite that should cover our bases. Need to still test on a school, and then will iterate. Likely will need to add more to the config and then will feed that into this in order to ensure we fit all custom schools. I also imagine when we refactor schools, we'll have a better idea of what else we'll need to add to this process.
- UPDATE: Tested this on a custom school and everything works! 

## context
- Custom schools need model cards.

## questions
- No questions at this time. This is just the first pass, and we'll continue to improve.


src/student_success_tool/reporting/sections/custom/attribute_sections.py
```diff
@@ -0,0 +1,148 @@
...
+                    timeframe_phrase = (
+                        f"in achieving {unit_str} required for graduation"
+                    )
+                elif unit == "year":
+                    timeframe_phrase = f"within {unit_str}"
+                elif unit in {"term", "semester"}:
+                    timeframe_phrase = f"within {unit_str}"
+                elif unit == "pct_completion":
+                    timeframe_phrase = f"at {unit_str}"
+                else:
```
Mesh-ach, line 64: I'm thinking do we want to strictly enforce expected units here, falling back to within X, Y might be a little odd, if someone specifies e.g. gpa.

src/student_success_tool/reporting/sections/custom/attribute_sections.py
```diff
@@ -0,0 +1,148 @@
...
+
+        except Exception:
+            LOGGER.error(
+                "Unable to retrieve student criteria configuration in config",
+                exc_info=True,
+            )
+            return f"{card.format.indent_level(1)}- Student criteria configuration was unavailable."
+
+    @registry.register("checkpoint_section")
+    def checkpoint():
```
Mesh-ach, line 130: I see only credit, year, term, and semester are handled. If unit is something unexpected e.g. "pct_completion" or other unexpected values, there’s no else branch, we should probably add an assertion or add an else statement

src/student_success_tool/reporting/sections/custom/evaluation_sections.py
```diff
@@ -0,0 +1,127 @@
...
+        section_text.append(perf_table_func())
+
+        evaluation_sections.append("\n\n".join(section_text))
+
+    # Performance across splits
+    for csv_path in split_artifacts:
+        if csv_path.startswith("metrics/") and csv_path.endswith("_splits.csv"):
+            title = "Performance across Splits"
+            table_func = make_metric_table(csv_path, title)
+            registry.register("performance_by_splits_metric_table")(table_func)
```
Mesh-ach, line 114: i think we might be silently overwriting with each iteration, i.e. re‑registering the same name, what do you think?

src/student_success_tool/reporting/sections/custom/attribute_sections.py
vishpillai123: You can't specify GPA actually. This is the importance of having our custom configs setup. The pydantic model will enforce only certain fields for this particular attribute. Otherwise, you wouldn't be able to load the config properly (anything outside of the project config will raise a validation error).

vishpillai123: Custom config project is here: https://github.com/datakind/student-success-tool/blob/develop/src/student_success_tool/configs/custom.py

src/student_success_tool/reporting/sections/custom/attribute_sections.py
vishpillai123: Only these are allowed given the custom project config. Under CheckpointConfig:
`unit: t.Literal["credit", "year", "term", "semester"]`.

src/student_success_tool/reporting/sections/custom/attribute_sections.py
```diff
@@ -0,0 +1,148 @@
...
+        ordinal function.
+        """
+        try:
+            unit = card.cfg.preprocessing.checkpoint.unit
+            value = card.cfg.preprocessing.checkpoint.value
+            base_message = "The model makes this prediction when the student has"
+            if unit == "credit":
+                return f"{base_message} earned {card.cfg.preprocessing.checkpoint.value} credits."
+            elif unit in {"year", "term", "semester"}:
+                unit_label = unit + ("s" if value != 1 else "")
```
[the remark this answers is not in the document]
vishpillai123: Yes, we can, though we don't technically need an else here. We could raise a error if it falls outside, but this should be taken care of by the configs, which enables modularity here.

vishpillai123:
hey Mesh, just responded to the comments. A lot of your questions are around validation that will be coming from the the project config. Let me know if you have any questions.

src/student_success_tool/reporting/sections/custom/attribute_sections.py
Mesh-ach: yeah okay, thank you for clarifying

PR #258, opened 2025-07-15
adding note to remember to handle pre-cohort courses in this notebook

nm3224:
## changes & context
We no longer save the data out in the data assessment portion so we need to ensure any additional data cleaning (ex. dropping pre-cohort courses) we do there is carried on into / reflected in the pre processing notebook as well

### DIFF
--- a/pipelines/pdp/institution_id/01-preprocess-data-TEMPLATE.py
+++ b/pipelines/pdp/institution_id/01-preprocess-data-TEMPLATE.py
@@ -133,6 +133,20 @@ df_student_terms.columns.tolist()
 # COMMAND ----------
 
 # MAGIC %md
+# MAGIC ### Handling Pre-Cohort Courses
+# MAGIC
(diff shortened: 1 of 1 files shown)


kaylawilding (approved):

pipelines/pdp/institution_id/01-preprocess-data-TEMPLATE.py
```diff
@@ -132,6 +132,24 @@
...
+# MAGIC ### Handling Pre-Cohort Courses
+# MAGIC
+# MAGIC Please rememeber to check with your schools during the data assessment call how they would like pre-cohort course records to be handled.
+# MAGIC
+
+# COMMAND ----------
+
+# We usually drop pre-cohort course records; If school requests otherwise, please set include_pre_cohort_courses in your config to TRUE, re-load the config, THEN run this cell!
+
+if cfg.preprocessing.include_pre_cohort_courses == False:
```
vishpillai123, line 145:
```suggestion
if not cfg.preprocessing.include_pre_cohort_courses:
```

PR #264, opened 2025-07-30
fix/typo & aliases for custom schools in model cards

vishpillai123:
## changes
- Applicable for PDP & Custom: Feature selection's incomplete threshold (missing data threshold) was not formatted correctly in the base model card class, which affects both PDP and custom schools. Before `incomplete_threshold=0.2` would show up as 0.2% in the model card, so I fixed it to show up as 20%.
-  Applicable for PDP & Custom: If `split_col` doesn't exist in the training data then this breaks the build step, so adding a raising of Validation error in case it doesn't exist.
- Applicable to only Custom: Aliases are showing up correctly in most places, but it did not show up correctly in the "Disparities by Student Group" section which shows the flagged FNR plots. I realized this wasn't showing up because I didn't create the override! It was still using the base bias section, which is why nothing broke but we need to create an override for the alias to show up (otherwise it's just "friendly case" formatting). So, for example if in the `cfg.student_group_aliases`, we have `ethnicity_ipeds = "Ethnicity"`, then this would only show up as "Ethnicity Ipeds" under this section and not the alias which is "Ethnicity". 

## context
- Mostly minor typo/wording changes, but important for optimal clarity in the model cards.

## questions
- These are three changes all grouped into one PR, but let me know if you have any concerns.

### DIFF
--- a/src/student_success_tool/reporting/model_card/base.py
+++ b/src/student_success_tool/reporting/model_card/base.py
@@ -198,7 +198,7 @@ class ModelCard(t.Generic[C]):
             "number_of_features": str(feature_count),
             "collinearity_threshold": str(fs_cfg.collinear_threshold),
             "low_variance_threshold": str(fs_cfg.low_variance_threshold),
-            "incomplete_threshold": str(fs_cfg.incomplete_threshold),
+            "incomplete_threshold": str(int(fs_cfg.incomplete_threshold)*100),
         }
 
     def get_model_plots(self) -> dict[str, str]:
--- a/src/student_success_tool/reporting/model_card/base.py
+++ b/src/student_success_tool/reporting/model_card/base.py
@@ -131,6 +131,11 @@ class ModelCard(t.Generic[C]):
         )
         self.training_data = self.modeling_data
         if self.cfg.split_col:
+            if self.cfg.split_col not in self.modeling_data.columns:
+                raise ValueError(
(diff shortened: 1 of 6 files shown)


nm3224:
all looks good but the incomplete threshold actually got me wondering if we should change that. I feel like >20% missing is still a lot of data (ex. like 30% missing is still 70% populated). Should we change this? Maybe like a 35-40% threshold?

I also realized a lot of times term_program_study is usually only ~50% populated b/c students simply haven't declared a term yet/don't have one in earlier terms... is this something we should somehow account for?

vishpillai123:
thanks for reviewing @nm3224! This is really interesting point, and I'm curious to discuss more. I do think >20% missing values still means we have a lot of data ~80% is a lot. But my concern is when we include features > 50% in the model without understanding exactly why the feature has so many missing values, it could create issues down the line. For example, if during inference that number jumps up to 70-80%, then that would require retraining.

I think including features with missing values makes sense since we have imputation in AutoML, but experimenting with the different AutoML platforms and with doing inferences, I've learned that features with missing values can present some problems. Also with inference, what we're currently doing is giving back the top 5 features and sometimes the feature value is missing or "nan". I actually wanted to bring this up in the team meeting since schools may not know exactly what "nan" means. We could just fill this in as "Missing", but I think doing more work up front in terms of handling missing values during data assessment (we currently do this for the most part), and then handling during feature selection will be helpful.

kaylawilding (approved):
Good catches on these

PR #267, opened 2025-07-30
feat/automate model selection

vishpillai123:
## changes
- Created a mean bias score for a model by aggregating bias scores across flags and normalizing by the number of valid comparisons (# of "no bias", "low", "medium", and "high" flags while leaving out "insufficient data" flags).
- The docstring for `aggregate_bias_scores` has a lot of detail on how we're calculating mean bias score and context for the design. 
- To summarize our bias score calculation briefly (if you have questions, please refer to the docstring), we use the FNR magnitude as a bias flag score then weigh each score using  flag weights, which are 1.0 for high, 0.75 for medium, and 0.25 for low. We then sum all weighed scores and then divide by the number of valid comparisons.
- Added "model selection" section in our templates where we re-rank the top runs after bias evaluation by using "test_recall", "test_roc_auc", and "test_bias_score_mean". I tried to keep the number of metrics in the final section simple since we are already sorting for 6 metrics before we evaluate bias (log_loss/recall/roc_auc for test and val). Though, I bumped up the number that we are evaluating in the template from 3 -> 5, so we can evaluate more runs and ensure we're grabbing the best run.
- Added `tomlkit` dependency so we can select a model, then update the config directly by updating the run_id and experiment_id. This dependency is needed because it allows us to write to .toml files while preserving formatting. 
- Added "update_metadata" function so we can update the config after selecting a model. 

## context
- Manually selecting a model and updating the config is a current pain point in our workflow.
- Once we can automatically select a model & update the config directly then we can now run retraining pipelines all the way from preprocessing -> training -> create preliminary shap values -> create model card.
- Previously we were stuck between the 2nd and 3rd step (training -> create preliminary shap values), since we had to choose a model manually.
- By including a bias score, we can effectively find the best model according to performance and bias in a “hands-off” approach when running the training notebook.
- This enables us to start creating retraining pipelines.

## questions
- This is a pretty big PR, so let me know if you have questions. I tested the model selection on two schools and both gave solid results as the best model. 
- Let me know if you have any suggestions or concerns for how we calculate bias score or how we are selecting the top model.


pipelines/pdp/institution_id/02-train-model-TEMPLATE.py
```diff
@@ -286,3 +286,35 @@
...
+
+# MAGIC %md
+# MAGIC # model selection
+
+# COMMAND ----------
+
+# Rank top runs again after evaluation for model selection
+selected_runs = modeling.evaluation.get_top_runs(
+    experiment_id,
+    optimization_metrics=[
```
nm3224, line 300: i like this! what made you choose recall and roc_auc as the two metrics? why not like f1 and log loss or something? or accuracy? just curious

src/student_success_tool/modeling/bias_detection.py
```diff
@@ -341,6 +355,75 @@ def flag_bias(
...
+    """
+    Create model bias score by aggregating bias flag scores, while accounting
+    for valid subgroup comparisons. A bias score for a flag is simply its FNR magnitude.
+
+    We also use flag weights to ensure we're accounting for statistical significance, which is baked
+    into each flag (e.g. low has a p-value less than 0.1; medium or high has p-values less than 0.01).
+    This design captures both magnitude and statistical strength of disparities.
+
+    Once we sum all of our weighted bias flag scores, we then normalize using the number of valid
+    comparisons, which is the sum of all "no bias", "low", "medium", and "high" bias flags, while
```
nm3224, line 371: what's the benefit of having the sum? TBH most of the time i just look to see if there's high or medium bias - and if it's not in the test set then i generally write it off - but if it is in the test set i check to see how drastic it really is then decide if we should re-train. I think it could be cool if we could incorporate into this like a message that says "high bias in test set > retraining advised" or smth like that that tells u the severity of the issue and what action should be taken

vishpillai123: We're doing this for each split, so test is calculated separately from validate or train. 

Sum is done for our final score, in which we then divide by the number of valid comparisons. So, overall, we get an "average" bias score for a model. Does that make sense?

nm3224:
overall nice job i really like this and am excited for automation + scaling!! woohoo

pipelines/pdp/institution_id/02-train-model-TEMPLATE.py
vishpillai123: Yeah so we can talk about this more as a team, but school usually care the most about recall generally-speaking. 

ROC AUC is included to understand how well the model is distinguishing positive vs. negative class. I use it when I'm selecting a model in order to look at the model's ranking/discrimination ability.

I'm open to adding log_loss to this, but we are already picking the top models to evaluate bias using log loss. The issue with adding too many metrics to this is that we are adding `test_bias_score_mean`, and adding more metrics dilutes how much we account for bias (we use balanced score for our ranking). Right now, we are looking roughly at 2/3 performance and 1/3 bias since we have three metrics in our final sorting.

pipelines/pdp/institution_id/02-train-model-TEMPLATE.py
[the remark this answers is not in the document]
vishpillai123: 1 & 2: Fair points, though as a reminder - we're sorting twice. 

The first, we looking at six metrics, so `test_recall`, `val_recall`, `test_log_loss`, `val_log_loss`, `test_roc_auc`, `val_roc_auc`. We could add test and val f1-score here to make sure we're not completely dismissing precision. What do you think? My main reason for not including is that F1-score and accuracy can potentially be skewed for imbalanced datasets, but I think it's still a good idea to add. 

Our second sort is after we run bias on the runs from the first sort, we do our final ranking on `test_recall`, `test_roc_auc`, and `test_bias_score_mean`. 

3. Yes good idea. I don't have the implementation in this PR, but we can add it another PR. I am planning on creating a metric (call it maybe `overfit_score`), in which we subtract val_log_loss and train_log_loss, and then try to minimize in our balanced score. So, we show preference to runs that have limited difference in log loss between training and validation. Runs that have a high amount of overfitting have a higher value for `overfit_score`. 
4. Yes, we're just using test metrics in our final selection of the model as this is the second sort that we do.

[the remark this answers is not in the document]
nm3224: sure! let's add it to team meeting agenda this week

src/student_success_tool/modeling/bias_detection.py
vishpillai123: Yeah I think we can do this once we establish what the baseline is across schools. The bias score mean will also help us compare schools to each other.

But after we gather all of our schools, we can come up with general thresholds.

pipelines/pdp/institution_id/02-train-model-TEMPLATE.py
```diff
@@ -286,3 +284,36 @@
...
+)
+logging.info("top run ids for model selection = %s", selected_runs)
+
+# Extract the top run
+top_run_name, top_run_id = next(iter(selected_runs.items()))
+logging.info(f"Selected top run: {top_run_name} & {top_run_id}")
+
+# COMMAND ----------
+
+# Update config with run and experiment ids
```
kaylawilding, line 314: very cool

kaylawilding (approved):
Looks great!

PR #268, opened 2025-08-04
feat/h2o

vishpillai123:
## changes
- Created first pass on H2O wrapper modules that can train, log via mlflow, load, and run inference on data using a similar process to the one we use currently with Azure Databricks AutoML.
- Added sklearn imputation pipeline that handles missing values appropriately prior to training via H2O. We impute missing values, and also add missing value flags in case missing is predictive. This is a critical step since H2O assumes a clean dataset without missing values. Tree models can handle missing values & linear models cannot, so with our imputation pipeline, we can now utilize linear models in H2O and we have improved interpretability.
- We run inference using H2O's predict contributions. This method is super fast, and by feeding in a background H2OFrame (such as the training data), this provides baseline SHAP values and makes our predictions more robust, contextual, & consistent.
- We can aggregate our SHAP values from their base feature names and produce our familiar top features output or SHAP summary plots (with or without missing value flags). The missing value flags are treated like one-hot encoding variables and their SHAP value is aggregated into the base feature.
- Added unit tests. Also, I needed to update the unit tests for the bias and inference files to correctly mock Mlflow calls, otherwise I was getting an error where no local mlflow experiment existed. By mocking the calls, we no longer need a local tmp experiment and the tests should work in any environment. 
- NOTE: I realized the random mlflow names like "blue-dolphin72" or whatever actually comes from mlflow and isn't a databricks feature. It's when you start an mlflow run without specifying a name. I kept it in our H2O implementation for a sense of normalcy since we're used to it with Databricks AutoML.

## context
- We need to transition away from Azure Databricks and H2O will help us get there. This implementation of H2O can be ran in GCP or Azure.
- H2O AutoML performance matches Azure AutoML Databricks.
- We train H2O, capture an experiment summary mlflow run, and then log the top 50 models (example of how it looks in DB UI, basically looks a lot like DB AutoML).
- Once we can get the modules merged, I'll then work on adding templates, updating the inference pipeline (adding a new task for H2O), and then also updating the model cards etc. This already fits seamlessly into most of our evaluation code (including bias evaluation).

## questions
- There's a lot happening here, let me know if you have questions.

image


vishpillai123:
@kaylawilding, hey so this is finally ready for review! Data type issue with "program_of_study_term_1" has been resolved. The issue was happening during imputation.

src/student_success_tool/modeling/h2o_modeling/training.py
```diff
@@ -0,0 +1,226 @@
...
+
+import h2o
+from h2o.automl import H2OAutoML
+
+from . import utils
+from . import imputation
+
+LOGGER = logging.getLogger(__name__)
+
+VALID_H2O_METRICS = {
```
kaylawilding, line 16: should f1 and/or recall be here?

src/student_success_tool/modeling/h2o_modeling/training.py
```diff
@@ -0,0 +1,226 @@
...
+    if client is None:
+        client = MlflowClient()
+
+    # Set and validate inputs
+    seed = kwargs.pop("seed", 42)
+    timeout_minutes = int(float(str(kwargs.pop("timeout_minutes", 5))))
+    exclude_cols = [
+        c for c in t.cast(list[str], kwargs.pop("exclude_cols", [])) if c is not None
+    ]
+    split_col: str = str(kwargs.pop("split_col", "split"))
```
kaylawilding, line 45: do we need do sample weight here? Can h20 do sample weights?

src/student_success_tool/modeling/h2o_modeling/training.py
```diff
@@ -0,0 +1,226 @@
...
+    processed_model_features = [c for c in train.columns if c not in exclude_cols]
+    LOGGER.info(f"Running H2O AutoML with {len(processed_model_features)} features...")
+
+    aml = H2OAutoML(
+        max_runtime_secs=timeout_minutes * 60,
+        sort_metric=metric,
+        stopping_metric=metric,
+        seed=seed,
+        verbosity="info",
+        include_algos=["XGBoost", "GBM", "GLM", "DRF"],
```
kaylawilding, line 131: do we want to hardcode this set of algorithms? or do we want to allow people to exclude some still? If not, which I think is probably fine (we may have some backwards compatibility issues on older schools but we can sort that out), we would need to update the config? actually maybe we can just make another config template for h20

src/student_success_tool/modeling/h2o_modeling/training.py
```diff
@@ -0,0 +1,226 @@
...
+    missing_cols = [c for c in exclude_cols if c not in df.columns]
+    if missing_cols:
+        raise ValueError(f"exclude_cols contains missing columns: {missing_cols}")
+
+    # Set training experiment
+    experiment_id = utils.set_or_create_experiment(
+        workspace_path=str(workspace_path),
+        institution_id=institution_id,
+        target_name=str(target_name),
+        checkpoint_name=str(checkpoint_name),
```
kaylawilding, line 81: just curious how the checkpoint name and target name are used here - this is setting the name of the experiment like we do now?

src/student_success_tool/modeling/h2o_modeling/training.py
```diff
@@ -0,0 +1,226 @@
...
+
+    aml = H2OAutoML(
+        max_runtime_secs=timeout_minutes * 60,
+        sort_metric=metric,
+        stopping_metric=metric,
+        seed=seed,
+        verbosity="info",
+        include_algos=["XGBoost", "GBM", "GLM", "DRF"],
+        nfolds=0,  # disable CV, use validation frame for early stopping
+        balance_classes=True,
```
kaylawilding, line 133: oh does this handle our sample weighting for us? seems like it might from h20s documentation - thatd be cool

src/student_success_tool/modeling/h2o_modeling/training.py
kaylawilding: oh i see these are just what we can use for our primary metric on but not for eval, but  i suppose we used to be able/currently can do these, right. So we will need to clarify in the new templates or something these arent options naymore.

src/student_success_tool/modeling/h2o_modeling/imputation.py
```diff
@@ -0,0 +1,437 @@
...
+
+
+class SklearnImputerWrapper:
+    """
+    A leakage-safe imputer using sklearn's SimpleImputer and ColumnTransformer,
+    with skew-aware numeric strategy assignment, optional missingness flags,
+    and MLflow-based artifact logging.
+    """
+
+    DEFAULT_SKEW_THRESHOLD = 0.5
```
kaylawilding, line 33: something we need to add as a new config param? do you think people would want to tweak this?

pipelines/pdp/institution_id/h2o/02-train-h2o-model-TEMPLATE.py
```diff
@@ -0,0 +1,333 @@
...
+logging.basicConfig(level=logging.INFO, force=True)
+logging.getLogger("py4j").setLevel(logging.WARNING)  # ignore databricks logger
+
+try:
+    spark = DatabricksSession.builder.getOrCreate()
+except Exception:
+    logging.warning("unable to create spark session; are you in a Databricks runtime?")
+    pass
+
+try:
```
kaylawilding, line 69: oh what is it doing with emails?

pipelines/pdp/institution_id/h2o/02-train-h2o-model-TEMPLATE.py
```diff
@@ -0,0 +1,333 @@
...
+# MAGIC %md
+# MAGIC # evaluate model
+
+# COMMAND ----------
+
+# HACK: Evaluate an experiment you've already trained
+# experiment_id = cfg.model.experiment_id
+
+# NOTE: AutoML generates a split column if not manually specified
+split_col = training_params.get("split_col", "_automl_split_col_0000")
```
kaylawilding: is this still relevant?

src/student_success_tool/modeling/h2o_modeling/utils.py
```diff
@@ -0,0 +1,407 @@
...
+        aml=aml,
+        leaderboard_df=leaderboard_df,
+        train=train,
+        valid=valid,
+        test=test,
+        target_col=target_col,
+    )
+
+    # Capping # of models that we're logging to save some time
+    MAX_MODELS_TO_LOG = 50
```
kaylawilding, line 109: lets move this as a default param or somewhere else a little more visible?

src/student_success_tool/modeling/h2o_modeling/utils.py
```diff
@@ -0,0 +1,407 @@
...
+                f"Completed logging on {model_num}/{len(top_model_ids)} top models..."
+            )
+
+        # Setting threshold to 0.5 due to binary classification
+        metrics = log_h2o_model(
+            model_id=model_id,
+            train=train,
+            valid=valid,
+            test=test,
+            threshold=0.5,
```
kaylawilding: maybe just leave it as using the default so we dont have 0.5 written in two places?

tests/modeling/h2o_modeling/test_inference.py
```diff
@@ -0,0 +1,138 @@
...
+import pytest
+from unittest import mock
+from student_success_tool.modeling.h2o_modeling import inference
+
+
+# Existing SHAP grouping test
+def test_group_shap_by_feature_basic():
+    df = pd.DataFrame(
+        {
+            "gender.M": [0.1, 0.2],
```
kaylawilding, line 11: does h20 use like var.value for one hot encoded cat vars names?

kaylawilding (approved):

src/student_success_tool/modeling/h2o_modeling/training.py
vishpillai123: I think H2O may not be able to do F1 or Recall, which is fine since we mostly just use log_loss

src/student_success_tool/modeling/h2o_modeling/training.py
vishpillai123: Yes exactly, it's used to set the name of the experiment similar to how we do it now.

src/student_success_tool/modeling/h2o_modeling/training.py
vishpillai123: Yeah good catch. So, I tried adding sample weight as a `weights_column`, which is a parameter in `H2OAutoML.train`. Unfortunately, when I extracted the features in the model using `names = list(model._model_json["output"]["names"][:-1])`, it came out as one of the features and when I threw this into `model.predict`, it gave an error. Like it's not being technically used by the model, but it's still showing up in the output JSON list, which is weird. We could technically filter it out, but basically, I don't know how `weights_column` is used exactly under the hood, so I think what we can do instead is either use `balance_classes=True` when we want sample weight and `balance_classes=False` when we don't want sample weight.

src/student_success_tool/modeling/h2o_modeling/training.py
vishpillai123: Yeah we definitely should hardcode the exact algos we need because H2O also has deep learning or stacked ensemble algorithms that we are trying to disable. We don't want to use those since we need interpretability.

This line of `include_algos` basically ensures that we only use tree or linear based models. We could technically add conditions where we don't include XGBoost or DRF (Random Forest) for our older schools. I'm curious to see how H2O treats these older schools though, but we could add logic for this in the function.

src/student_success_tool/modeling/h2o_modeling/training.py
vishpillai123: Yeah, we use it to set the experiment name.

src/student_success_tool/modeling/h2o_modeling/training.py
vishpillai123: Yes, it handles sample weight in this way. I think it has its own internal algorithm of whether the models truly need sample weight, but I need to test this out.

pipelines/pdp/institution_id/h2o/02-train-h2o-model-TEMPLATE.py
vishpillai123: We need the workspace email in order to set the experiment name automatically. Databricks AutoML does this under the hood.

pipelines/pdp/institution_id/h2o/02-train-h2o-model-TEMPLATE.py
vishpillai123: Nah not relevant - I'll remove it.

src/student_success_tool/modeling/h2o_modeling/training.py
vishpillai123: I need to test out how H2O AutoML fairs with class imbalance in general honestly.

src/student_success_tool/modeling/h2o_modeling/utils.py
vishpillai123: We can, but I think that logging more than the top 50 models is not necessary. Logging via mlflow takes a long time. When I train a model with 10 minute timeout, it will finish training and then take an extra 6-7 mins to log the top 50 models.. logging more than that will take even longer.

We only need 5-10 top models for model evaluation and selection, and H2O AutoML will create hundreds of models in 10 mins (it's super fast). 

What concerns do you have with logging only top 50?

tests/modeling/h2o_modeling/test_inference.py
vishpillai123: Yeah it uses var.value, which is super convenient for inference since we later collapse the SHAP values using this.

src/student_success_tool/modeling/h2o_modeling/utils.py
vishpillai123: yes, good move - I can do that.

src/student_success_tool/modeling/h2o_modeling/utils.py
vishpillai123: Also I'm still saving the number of models that H2O trained (which is used for reporting) in `log_h2o_experiment_summary`, which creates a run called "H2O AutoML Experiment Summary and Storage" (similar to databricks's "Training Data Storage and Analysis" run)

PR #278, opened 2025-08-07
minor bug in feature table & linting

nm3224:
## changes & context

- The feature has a "has" during creation so these features aren't being mapped right
- adding "so far" instead of just "so"
- adding "from databricks.sdk.runtime import dbutils" to template for linting error related to dbutils


pipelines/pdp/institution_id/02-train-model-TEMPLATE.py
```diff
@@ -41,6 +41,7 @@
 import logging
 import mlflow
 from databricks.connect import DatabricksSession
+from databricks.sdk.runtime import dbutils
```
vishpillai123, line 44: did we mean to add this?

src/student_success_tool/assets/pdp/features_table.toml
```diff
@@ -254,8 +254,8 @@ section_num_students_enrolled_std = { name = "standard deviation in students enr
...
+student_has_earned_certificate_at_cohort_inst = { name = "student has received a certificate from cohort institution", long_desc = "A binary feature (True/False) indicating whether or not a student has earned a certificate from the cohort institution before enrolling in their current program. A value of \"True\" means the student completed a certificate program at the cohort institution, and is now enrolled in an additional program, while a value of \"False\" suggests no such prior certificate earned at the cohort institution.", short_desc = "Whether or not a student previously earned a certificate from the cohort institution before enrolling in their current program" }
```
vishpillai123, line 257: how do we know for sure that it "has" is in the feature name?

pipelines/pdp/institution_id/02-train-model-TEMPLATE.py
nm3224: yea it's for linting so u don't get an error "dbutils not found"

src/student_success_tool/assets/pdp/features_table.toml
nm3224: when mapping, it's not being mapped correctly and i checked the feature creation, it has "has" in it

pipelines/pdp/institution_id/02-train-model-TEMPLATE.py
vishpillai123: got it, usually we can `# noqa: F821`, which helps us ignore lint whenever we use dbutils, but this is fine too.

src/student_success_tool/assets/pdp/features_table.toml
vishpillai123: Ok, I wonder how we can ensure that the feature table always has a mapping to features. Is that possible?

vishpillai123 (approved):

src/student_success_tool/assets/pdp/features_table.toml
nm3224: maybe we can add to unit tests?

PR #280, opened 2025-08-21
template updates

nm3224:
## changes & context 

- adding in course level patterns we agreed on as a default
- making 3 diff configs for retention, credits_earned and graduation where we assume graduation is usually 150% time for associate's degree at a 30 credit checkpoint or term 2 only core terms (added TODO to select one of these and remove other)
- remove any default course ids or cip codes incase of user error of forgetting to change them, better to have them empty
- changed config reading in 00 data assessment notebook to remove "template"
- Adding TODO note for the removal of pre cohort courses (or not)
- having the default retention config ready will help us for our future "default model" planning too
- i don't think we need the "mangled column names" converter func anymore so removing that
- adding converter func to drop real duplicates

## questions
- how can i get rid of the unit test error? 
- we should auto-populate the max academic year based on the data-how can we update this?


pipelines/pdp/institution_id/00-data-assessment-TEMPLATE.py
```diff
@@ -78,7 +78,7 @@
 # COMMAND ----------
 
 # project configuration should be stored in a config file in TOML format
-cfg = dataio.read_config("./config-TEMPLATE.toml", schema=configs.pdp.PDPProjectConfig)
+cfg = dataio.read_config("./config.toml", schema=configs.pdp.PDPProjectConfig)
```
kaylawilding, line 81: oh we should make this is a TODO now since we have multiple configs

src/student_success_tool/configs/pdp.py
```diff
@@ -136,46 +136,6 @@ def mlflow_model_uri(self) -> str:
         return f"runs:/{self.run_id}/model"
```
kaylawilding: did you need to move this block?

tests/configs/test_pdp.py
```diff
@@ -11,16 +11,24 @@
...
+    "config-RETENTION_TEMPLATE.toml",
+]
 
-@pytest.fixture(scope="module")
-def template_cfg_dict():
-    config_path = SRC_ROOT / "config-TEMPLATE.toml"
-    with config_path.open("rb") as f:
-        return tomllib.load(f)
 
+@pytest.mark.parametrize("config_file", CONFIG_FILES)
```
kaylawilding, line 21: this looks good and I do think was necessary

nm3224:
@kaylawilding do we want to merge this?

src/student_success_tool/configs/pdp.py
```diff
@@ -343,6 +303,57 @@ class TargetCreditsEarnedConfig(TargetBaseConfig):
     max_term_rank: int | t.Literal["infer"]
 
 
+class PreprocessingConfig(pyd.BaseModel):
```
vishpillai123: was this just moved to a different part of the file?

tests/configs/test_pdp.py
```diff
@@ -11,16 +11,24 @@
 
 SRC_ROOT = pathlib.Path(__file__).parents[2] / "pipelines" / "pdp" / "institution_id"
 
+CONFIG_FILES = [
```
vishpillai123, line 14: nice job with adding this in for the unit test!!

src/student_success_tool/configs/pdp.py
nm3224: yes, i think b/c the later funcs call this and so it needs to exist first

tests/configs/test_pdp.py
nm3224: thanks!

src/student_success_tool/configs/pdp.py
vishpillai123: oh but this was working before? If we didn't change any of the logic of the config, I think it's okay to leave it as is.

nm3224: no it wasn't b/c i changed some of the unit tests too

src/student_success_tool/configs/pdp.py
```diff
@@ -343,6 +303,57 @@ class TargetCreditsEarnedConfig(TargetBaseConfig):
...
+    features: "FeaturesConfig"
+    selection: "SelectionConfig"
+    checkpoint: t.Union[
+        CheckpointNthConfig,
```
nm3224: @vishpillai123  this is different

src/student_success_tool/configs/pdp.py
```diff
@@ -343,6 +303,57 @@ class TargetCreditsEarnedConfig(TargetBaseConfig):
...
+        CheckpointFirstConfig,
+        CheckpointLastConfig,
+        CheckpointFirstAtNumCreditsEarnedConfig,
+        CheckpointFirstWithinCohortConfig,
+        CheckpointLastInEnrollmentYearConfig,
+    ]
+    target: t.Union[
+        TargetGraduationConfig,
```
nm3224: @vishpillai123  and this

src/student_success_tool/configs/pdp.py
vishpillai123: The type union? The checkpointNthconfig existed before.

Not sure if type check was being goofy and you had to add this.

src/student_success_tool/configs/pdp.py
vishpillai123: oh ok, thanks for clarifying.

vishpillai123 (approved):

src/student_success_tool/dataio/converters/pdp/raw_course.py
```diff
@@ -70,3 +70,40 @@ def dedupe_by_renumbering_courses(df: pd.DataFrame) -> pd.DataFrame:
     # update course numbers for dupe'd records in-place
     df.update(deduped_course_numbers, overwrite=True)
     return df
+
+
+def drop_real_duplicates(df: pd.DataFrame) -> pd.DataFrame:
```
kaylawilding, line 75: did we decide to keep this?  I suppose since we need to do other src edits anyhow we can keep it

kaylawilding (approved):

PR #282, opened 2025-08-22
streamlining H2O modules, templates, and adding sample weight

vishpillai123:
This is a huge PR (I'm sorry in advance).

Updates:
- Sample weight: Looks like we do need sample weight column in order to trigger sample weight during training (balance classes basically does nothing lol). So, I added this back in and then ensured that when we grab features from the model that we filter out this column. For whatever reason, H2O includes `weights_column` in the features even though it's not technically a feature.
- Inference: Added background data sampling to templates.
- Model Registration: In order for us to register models , Unity catalog requires a model signature, `conda.yaml`, and `requirements.txt`. I created a super lightweight version of this so that can register H2O models in UC.
- Imputation: Ensured that excluded columns from fit (student_id, target) are not filtered out during transform, so that we don't have to deattach and re-attach. It leaves those columns untouched and only imputes columns that were part of the sklearn imputer fit.
- Configs: Added H2O separate project configs, unit tests, etc. The configs are basically the same honestly... "log_loss" is just referred to as "logloss" within H2O. I removed sklearn framework and added a background_data_sample. 
- Model Cards: Added H2O separate model cards.

Other smaller updates:
- Added a `_to_pandas(..)` conversion function which uses multi-threading to convert H2OFrame into a pandas Dataframe.
- Added a `_to_h2o(..)` conversion function that converts pandas to h2o but also (more importantly) corrects data types post h2o conversion. H2O will sometimes infer dtypes incorrectly (for example it converts program of study term which is usually "1345.0" as an int, so we have to revert it back to enum which is a category type in h2o). We use the original dataframe to know how to correct dtypes. 
- Basically, we're abstracting any pandas -> h2o or h2o -> pandas away from the user, which makes things streamlined and standardized. We also log everything, which will also help in data drift scenarios.
- Added ability to exclude classifiers such as XGBoost or GBM from training (this is for backward compatibility with our older schools).
- Created an H2O `model_comparison.png` plot. Databricks AutoML creates this in the backend, and we use it for model cards so I had to create an H2O version.


vishpillai123:
hi @kaylawilding, this is ready for review! Let me know if you have any questions.

pipelines/custom/institution_id/h2o/config-TEMPLATE.toml
```diff
@@ -0,0 +1,121 @@
...
+predict_table_path = "CATALOG.INST_ID_silver.DATASET_NAME_preprocessed_inference"
+
+[datasets.silver.course]
+train_table_path = "CATALOG.INST_ID_silver.DATASET_NAME_course_train"
+predict_table_path = "CATALOG.INST_ID_silver.DATASET_NAME_course_inference"
+
+[datasets.silver.semester]
+train_table_path = "CATALOG.INST_ID_silver.DATASET_NAME_semester_train"
+predict_table_path = "CATALOG.INST_ID_silver.DATASET_NAME_semester_inference"
+
```
kaylawilding, line 53: I think youre missing datasets.silver.preprocessed or at least it still used in the h20 template

pipelines/custom/institution_id/h2o/config-TEMPLATE.toml
```diff
@@ -0,0 +1,121 @@
...
+[preprocessing.target]
+name = "graduation_150pct_time"
+category = "graduation"
+unit = "pct_completion"
+value = 150
+
+[preprocessing.target.params]
+intensity_time_limits = { FULL-TIME = [3.0, "year"], PART-TIME = [6.0, "year"] }
+
```
kaylawilding, line 102: we need pos_label somewhere?

pipelines/custom/institution_id/h2o/03-make-h2o-predictions-TEMPLATE.py
```diff
@@ -44,9 +40,10 @@
 from student_success_tool import configs, dataio, modeling
 from student_success_tool.modeling import h2o_modeling
 
+
```
kaylawilding, line 43: need 
from py4j.protocol import Py4JJavaError

kaylawilding (approved):

kaylawilding:
Just two little template things that arent a big deal - we can get them in the next merge. Going to keep testing this on PDP schools and we can tweak dtype stuff if we need in the next merge

vishpillai123:
> Just two little template things that arent a big deal - we can get them in the next merge. Going to keep testing this on PDP schools and we can tweak dtype stuff if we need in the next merge

sounds good! I'll also suppress the output when logging... i think i missed adding the wrapper on one of the log functions! I'll create a quick PR for that next week along with the template changes.

PR #284, opened 2025-08-26
changing config

nm3224:
### Config Changes: 
- adding default filtersfor retention config
      - core terms only (can be adjusted) 
      - associate's students only
      - first time admits only 
- changing incomplete threshold from 0.5 to 0.2
- added extra parameters with a TODO to edit based on school's preferences for checkpoint; default is always core terms only and excluding any pre-cohort terms _unless specified otherwise_
     - exclude_non_core_terms = true
     - exclude_pre_cohort_terms = true

### Template Changes:
- adding gateway course cip codes as a default too in the template
- adding prints for top 10 and bottom 10 support scores for slide deck visibility
- adding catalog variable so only have to change in one place 

### Checkpointing Logic Changes:
- changing checkpointing code for retention to default as excluding non-core terms and pre-cohort terms alike
- adding descriptive sentences to args missing them

### Feature Table Changes:
- adding "number of" at the start and "taken this term" at the end


pipelines/pdp/institution_id/config-CREDITS_EARNED_TEMPLATE.toml
```diff
@@ -65,7 +65,7 @@ num_terms_in_year = 4
 max_term_rank = "infer"
 
 [modeling.feature_selection]
-incomplete_threshold = 0.5
+incomplete_threshold = 0.8
```
kaylawilding: We want to drop features with 20% or more missing correct? The function that uses this drops features with greater than or equal to the threshold missing values. I think this should be 0.2. Double check the fn defn for drop_incomplete_features in feature selection though please. Also, we should probably change the default in that function as well from 0.5. If Im understanding this correctly, please make the same edit to the other config templates

pipelines/pdp/institution_id/config-RETENTION_TEMPLATE.toml
```diff
@@ -42,7 +42,8 @@ key_course_subject_areas = []
 key_course_ids = []
 
 [preprocessing.selection]
-student_criteria = { enrollment_type = "", credential_type_sought_year_1 = "" }
+# default is usually fall-to-fall retention, if otherwise then change to core_terms
+student_criteria = { enrollment_type = "", credential_type_sought_year_1 = "", cohort_term = "FALL" }
```
kaylawilding: Lets set this up for the default models? We can chat about what should the default model params be in our 1-1 if that helps

pipelines/pdp/institution_id/03-make-predictions-TEMPLATE.py
```diff
@@ -273,12 +279,14 @@
 
 with mlflow.start_run() as run:
     confusion_matrix = evaluation.log_confusion_matrix(
+        catalog="sst_dev",
```
kaylawilding: actually can you add a variable for catalog and then make these four instance call that var - just so we only have to edit one place instead of 4 if it needs changed?

pipelines/pdp/institution_id/01-preprocess-data-TEMPLATE.py
```diff
@@ -135,6 +136,11 @@
 
 # COMMAND ----------
 
+# confirm cip code list, then add to config
+cip_code_list = ["23", "27"]
```
kaylawilding, line 140: sorry whats happening here? these are the cip codes for the math and english gateways? the user needs to confirm these are correct for the school and then add to the config?

kaylawilding (changes requested):

pipelines/pdp/institution_id/config-CREDITS_EARNED_TEMPLATE.toml
nm3224: oh shoot, i thought it was we want to drop features with 80% or more missing... @vishpillai123 am i understanding this wrong

nm3224: @kaylawilding i'll check on this and change as needed!

pipelines/pdp/institution_id/config-RETENTION_TEMPLATE.toml
nm3224: sounds good!

pipelines/pdp/institution_id/03-make-predictions-TEMPLATE.py
nm3224: yup!

pipelines/pdp/institution_id/01-preprocess-data-TEMPLATE.py
nm3224: yes correct!

nm3224: it typically is 23 and 27 for like all the schools i've worked on, but user shud always confirm

pipelines/pdp/institution_id/config-CREDITS_EARNED_TEMPLATE.toml
nm3224: changed! I think you're right @kaylawilding

src/student_success_tool/preprocessing/checkpoints/pdp.py
```diff
@@ -190,15 +190,18 @@ def first_student_terms_at_num_credits_earned(
...
     df: pd.DataFrame,
     *,
-    term_is_pre_cohort_col: str = "term_is_pre_cohort",
     student_id_cols: str | list[str] = "student_id",
     sort_cols: str | list[str] = "term_rank",
     include_cols: t.Optional[list[str]] = None,
+    term_is_pre_cohort_col: str = "term_is_pre_cohort",
+    exclude_pre_cohort_terms: bool = True,
+    term_is_core_col: str = "term_is_core",
+    exclude_non_core_terms: bool = True,
```
nm3224, line 199: add to config

src/student_success_tool/preprocessing/checkpoints/pdp.py
```diff
@@ -190,15 +190,18 @@ def first_student_terms_at_num_credits_earned(
 def first_student_terms_within_cohort(
     df: pd.DataFrame,
     *,
-    term_is_pre_cohort_col: str = "term_is_pre_cohort",
     student_id_cols: str | list[str] = "student_id",
     sort_cols: str | list[str] = "term_rank",
     include_cols: t.Optional[list[str]] = None,
+    term_is_pre_cohort_col: str = "term_is_pre_cohort",
```
nm3224, line 196: check if this is redundant

pipelines/pdp/institution_id/03-make-predictions-TEMPLATE.py
```diff
@@ -308,7 +319,7 @@
 # save sample advisor output dataset
 dataio.write.to_delta_table(
     shap_feature_importance,
-    f"staging_sst_01.{cfg.institution_id}_silver.training_{cfg.model.run_id}_shap_feature_importance",
+    f"sst_dev.{cfg.institution_id}_silver.training_{cfg.model.run_id}_shap_feature_importance",
```
vishpillai123: can you add catalog here as well?

pipelines/pdp/institution_id/03-make-predictions-TEMPLATE.py
```diff
@@ -328,6 +339,6 @@
 # save sample advisor output dataset
 dataio.write.to_delta_table(
     support_score_distribution,
-    f"staging_sst_01.{cfg.institution_id}_silver.training_{cfg.model.run_id}_support_overview",
+    f"sst_dev.{cfg.institution_id}_silver.training_{cfg.model.run_id}_support_overview",
```
vishpillai123: and catalog here as well

pipelines/pdp/institution_id/03-make-predictions-TEMPLATE.py
nm3224: done!

vishpillai123 (approved):

pipelines/pdp/institution_id/01-preprocess-data-TEMPLATE.py
```diff
@@ -239,6 +245,7 @@
 
 # TODO: choose checkpoint function suitable for school's use case
 # parameters should be specified in the config
+# MAKE SURE TO SPECIFY exclude_non_core_terms and exclude_pre_cohort_terms per school's needs!
```
kaylawilding, line 248: 👍

pipelines/pdp/institution_id/config-RETENTION_TEMPLATE.toml
```diff
@@ -42,19 +42,23 @@ key_course_subject_areas = []
 key_course_ids = []
 
 [preprocessing.selection]
-student_criteria = { enrollment_type = "", credential_type_sought_year_1 = "" }
+# if fall-to-fall retention, change cohort_term = "FALL"
+student_criteria = { enrollment_type = "FIRST-TIME", credential_type_sought_year_1 = "ASSOCIATE'S DEGREE", cohort_term = ["FALL", "SPRING"] }
```
kaylawilding, line 46: 👍

src/student_success_tool/preprocessing/checkpoints/pdp.py
```diff
@@ -190,15 +190,18 @@ def first_student_terms_at_num_credits_earned(
 def first_student_terms_within_cohort(
```
kaylawilding, line 190: This changes the default from what older schools would have used, correct? Can you just put this on your list to document in confluence this week?

nm3224: yes- they would just have to set `exclude_non_core_terms` to false for it to be the same as before (we were alr removing pre-cohort anyway)

nm3224: @kaylawilding where in the confluence should the change go / shud i slack the team ab this in our channel

src/student_success_tool/preprocessing/targets/pdp/graduation.py
```diff
@@ -117,10 +117,11 @@ def compute_target(
         df,
         checkpoint=ft.partial(
             checkpoints.pdp.first_student_terms_within_cohort,
-            term_is_pre_cohort_col=term_is_pre_cohort_col,
             student_id_cols=student_id_cols,
             sort_cols=term_rank_col,
             include_cols=(student_id_cols + [term_rank_col, enrollment_intensity_col]),
+            term_is_pre_cohort_col=term_is_pre_cohort_col,
+            term_is_core_col="term_is_core",
```
kaylawilding, line 124: I think ideally in the long-run lets not hardcode this feature but we can put that on the backlog

kaylawilding (approved):

pipelines/pdp/institution_id/03-make-predictions-TEMPLATE.py
```diff
@@ -308,7 +319,7 @@
 # save sample advisor output dataset
 dataio.write.to_delta_table(
     shap_feature_importance,
-    f"staging_sst_01.{cfg.institution_id}_silver.training_{cfg.model.run_id}_shap_feature_importance",
+    f"catalog.{cfg.institution_id}_silver.training_{cfg.model.run_id}_shap_feature_importance",
```
vishpillai123: this needs to be {catalog} since it's an f-string

src/student_success_tool/preprocessing/checkpoints/pdp.py
kaylawilding: ya just slack me or I think I made you an asana task - just tag me there please :)

pipelines/pdp/institution_id/03-make-predictions-TEMPLATE.py
kaylawilding: oh good catch

PR #285, opened 2025-09-01
Box whispers plot

Mesh-ach:
Add feature_boxstats: five-number summary for global top-N SHAP features (box-and-whisker ready)

## changes
- Added feature_boxstats helper

## context
- Function to feed endpoint for FE box plot

## questions
No questions at this time


src/student_success_tool/modeling/inference.py
```diff
@@ -412,6 +412,82 @@ def top_shap_features(
...
+        raise ValueError(
+            f"shap_values shape {shap_values.shape} must match features shape {features.shape}"
+        )
+
+    mean_abs = pd.Series(np.mean(np.abs(shap_values), axis=0), index=features.columns)
+    top_feats = mean_abs.sort_values(ascending=False)
+
+    # Restrict stats to numeric columns
+    stats_source = (
+        features.select_dtypes(include=[np.number]) if numeric_only else features
```
vishpillai123: If `numeric_only` is False, I think col.min(), col.quantile(...), and col.max() would throw errors right?

I think we can remove that as a parameter and just assume that we are always working with numeric features.

vishpillai123 (approved):

PR #288, opened 2025-09-03
changed ifnerence table paths to use databricks run id instead of mod…

Mesh-ach:
Changed inference FE table names to use databricks run id

## changes
- I just realised the inference table names are still using model run ids, instead of the databricks run id as decided. I corrected that in this PR


vishpillai123:
@Mesh-ach thanks for your initiative on this!! Closing since this is covered in Kayla's updates to the inference pipeline here: https://github.com/datakind/student-success-tool/pull/290

PR #289, opened 2025-09-06
feat/h2o templates cleanup

vishpillai123:
**compute_dataset_splits**: Added stratification. Also added using train_test_split because the performance improvement of using np.rng is only marginal. In fact using np.rng as we have been was actually a bug since we were all using the same seed and target prevalence was more in validation than test since we weren't using stratification. Especially with smaller datasets, this difference becomes magnified. In H2O, this caused validation log loss to always be artifically greater than test log loss (because it has more 1s than 0s compared to test log loss).

Also updated preprocess notebook to add the stratify column, which in our case is just our target column.

**H2O training configuration**: Removed stopping threshold and rounds since it can cause H2O training to be much slower and the performance improvement isn't a lot. Also, added nfolds to just be set to 5 which should be applicable for all of our datasets since we are mostly always under 50K rows. 

**templates**: Made some small import changes based on the last PR.

**H2O progress bar disabled**: Disabled progress bar which cleans up logs and makes things more readable.

### DIFF
--- a/src/student_success_tool/modeling/h2o_modeling/utils.py
+++ b/src/student_success_tool/modeling/h2o_modeling/utils.py
@@ -13,8 +13,12 @@ from mlflow.models import Model, infer_signature
 from mlflow.tracking import MlflowClient
 import pandas as pd
 import numpy as np
-from pandas.api.types import is_categorical_dtype, is_object_dtype, is_string_dtype
-
+from pandas.api.types import (
+    is_categorical_dtype,
+    is_object_dtype,
+    is_string_dtype,
+    is_integer_dtype,
+)
 
 import h2o
 from h2o.automl import H2OAutoML
@@ -58,6 +62,8 @@ def safe_h2o_init(base_port: int = 54321, mem_per_cluster: str = "4G") -> None:
         LOGGER.warning(
             "H2O is not bound to localhost. This may expose REST endpoints on the network."
         )
+    # Stop H2O progress bars globally
+    h2o.no_progress()
 
 
 def download_artifact_file(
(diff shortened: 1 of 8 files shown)


src/student_success_tool/modeling/h2o_modeling/utils.py
```diff
@@ -347,13 +353,14 @@ def log_h2o_model(
...
 
-            # params + metrics (use the batched version you implemented)
-            log_model_metadata_to_mlflow(
-                model_id=model_id,
-                model=model,
-                metrics=metrics,
-                exclude_keys={"model_id"},
-            )
+            with _suppress_output():
+                # params + metrics (use the batched version you implemented)
```
kaylawilding, line 357: 👍

src/student_success_tool/modeling/utils.py
```diff
@@ -21,38 +21,68 @@
...
-    fracs = list(label_fracs.values())
-    rng = np.random.default_rng(seed=seed)
-    return pd.Series(
-        data=rng.choice(labels, size=len(df), p=fracs, shuffle=shuffle),
-        index=df.index,
-        dtype="string",
-        name="split",
+    # Normalize fractions
+    total = sum(label_fracs.values())
+    fracs = {k: v / total for k, v in label_fracs.items()}
```
kaylawilding, line 51: should we check the fractions sum to ~1?

kaylawilding (approved):

src/student_success_tool/modeling/utils.py
vishpillai123: we actually validate this already in the project config!

PR #291, opened 2025-09-24
Fix inference shap importance

Mesh-ach:
added fix for features metadata

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1211364882184478


vishpillai123 (commented):
hey @Mesh-ach, can we instead handle this inside the function based on metadata being True? Otherwise, I feel like we may get inconsistency with how the short desc amd long desc get added from table to table in our pipeline. This what was also what created the void columns and corrupt tables in dev.

Also we would need to add to have this reflected  in our new repo: https://github.com/datakind/edvise/tree/develop

vishpillai123:
Also based on Caroline's latest message in the Asana task, looks like this is no longer needed? Can we confirm @kaylawilding ?

kaylawilding:
@vishpillai123 We do still need this

@Mesh-ach ya is this possible to pull inside the generate_ranked_feature_table function? If you think it can be, I can pull it into the new repo and test it real quick (no need to update this repo here)

Mesh-ach:
> Also based on Caroline's latest message in the Asana task, looks like this is no longer needed? Can we confirm @kaylawilding ?

yeah we still do Vish, sorry i linked the wrong ticket,

Mesh-ach:
> @vishpillai123 We do still need this
> 
> @Mesh-ach ya is this possible to pull inside the generate_ranked_feature_table function? If you think it can be, I can pull it into the new repo and test it real quick (no need to update this repo here)

I'm on it

vishpillai123:
@Mesh-ach @kaylawilding  Ok got it thanks for confirming. If that's the case, for the new repo can we make sure that for the training pipeline, we set metadata to False? At least when we're logging the artifact with mlflow? 

Otherwise, the tables with the short desc & long desc will for sure blow up the model cards. I even prefer for us to create a separate function for building the table.

Mesh-ach:
> @Mesh-ach @kaylawilding Ok got it thanks for confirming. If that's the case, for the new repo can we make sure that for the training pipeline, we set metadata to False? At least when we're logging the artifact with mlflow?
> 
> Otherwise, the tables with the short desc & long desc will for sure blow up the model cards. I even prefer for us to create a separate function for building the table.

So just to confirm, we’re should not change generate_ranked_function itself, just create a new one to generate the tables? That works for me, just wanted to make sure I’m following.

vishpillai123:
@Mesh-ach yeah this is confusing. Okay from what I understand here's what we need:

1. Ranked features table as an artifact (used in model card) - only feature name, data type, and average SHAP magnitude.
2. Training SHAP features FE table - do we need short desc and long desc here? Otherwise it'll be the same table from (1).
3. Inference SHAP features FE table - feature name, short desc, long desc, data type, average SHAP magnitude.

I built this function called `build_and_log_ranked_feature_table` that is used by both the training (with mlflow logging) & inference pipelines (without mlflow logging). https://github.com/datakind/edvise/blob/develop/src/edvise/scripts/predictions_h2o.py


@kaylawilding can you confirm whether the training FE table needs the same columns as inference?

Mesh-ach:
> @Mesh-ach yeah this is confusing. Okay from what I understand here's what we need:
> 
> 1. Ranked features table as an artifact (used in model card) - only feature name, data type, and average SHAP magnitude.
> 2. Training SHAP features FE table - do we need short desc and long desc here? Otherwise it'll be the same table from (1).
> 3. Inference SHAP features FE table - feature name, short desc, long desc, data type, average SHAP magnitude.
> 
> I built this function called `build_and_log_ranked_feature_table` that is used by both the training (with mlflow logging) & inference pipelines (without mlflow logging). https://github.com/datakind/edvise/blob/develop/src/edvise/scripts/predictions_h2o.py
> 
> @kaylawilding can you confirm whether the training FE table needs the same columns as inference?

yes, per discussion with Larry yesterday, I think we need the short desc and long desc for both training and inference tables

Mesh-ach:
> @Mesh-ach yeah this is confusing. Okay from what I understand here's what we need:
> 
> 1. Ranked features table as an artifact (used in model card) - only feature name, data type, and average SHAP magnitude.
> 2. Training SHAP features FE table - do we need short desc and long desc here? Otherwise it'll be the same table from (1).
> 3. Inference SHAP features FE table - feature name, short desc, long desc, data type, average SHAP magnitude.
> 
> I built this function called `build_and_log_ranked_feature_table` that is used by both the training (with mlflow logging) & inference pipelines (without mlflow logging). https://github.com/datakind/edvise/blob/develop/src/edvise/scripts/predictions_h2o.py
> 
> @kaylawilding can you confirm whether the training FE table needs the same columns as inference?

I think I’ve been working off an outdated branch (sst - develop), so I wasn’t aware of the changes made to the process so far.

vishpillai123:
> > @Mesh-ach yeah this is confusing. Okay from what I understand here's what we need:
> > 
> > 1. Ranked features table as an artifact (used in model card) - only feature name, data type, and average SHAP magnitude.
> > 2. Training SHAP features FE table - do we need short desc and long desc here? Otherwise it'll be the same table from (1).
> > 3. Inference SHAP features FE table - feature name, short desc, long desc, data type, average SHAP magnitude.
> > 
> > I built this function called `build_and_log_ranked_feature_table` that is used by both the training (with mlflow logging) & inference pipelines (without mlflow logging). https://github.com/datakind/edvise/blob/develop/src/edvise/scripts/predictions_h2o.py
> > @kaylawilding can you confirm whether the training FE table needs the same columns as inference?
> 
> yes, per discussion with Larry yesterday, I think we need the short desc and long desc for both training and inference tables

Ok sweet - then @Mesh-ach , can you take that function I built build_and_log_ranked_feature_table, and separate it so we have: 

1. One function that builds the 3-column table (feature name, data type, and average SHAP magnitude). This is called at training and inference time.
2. A separate function (only used at training) that logs the 3-column table as an artifact. It needs to be named in the same way it is now: `selected_features/ranked_selected_features.csv`. Otherwise model card won't be able to retrieve it.
3. Then the third function uses that 3-column table then adds the two other columns (short desc and long desc). This will be called at both training and inference. This 5 column table will be the FE table at training and inference.

Let me know if this implementation works for you or if you have a more DRY approach, feel free to do that. @kaylawilding to confirm above looks right.

Mesh-ach:
> > > @Mesh-ach yeah this is confusing. Okay from what I understand here's what we need:
> > > 
> > > 1. Ranked features table as an artifact (used in model card) - only feature name, data type, and average SHAP magnitude.
> > > 2. Training SHAP features FE table - do we need short desc and long desc here? Otherwise it'll be the same table from (1).
> > > 3. Inference SHAP features FE table - feature name, short desc, long desc, data type, average SHAP magnitude.
> > > 
> > > I built this function called `build_and_log_ranked_feature_table` that is used by both the training (with mlflow logging) & inference pipelines (without mlflow logging). https://github.com/datakind/edvise/blob/develop/src/edvise/scripts/predictions_h2o.py
> > > @kaylawilding can you confirm whether the training FE table needs the same columns as inference?
> > 
> > 
> > yes, per discussion with Larry yesterday, I think we need the short desc and long desc for both training and inference tables
> 
> Ok sweet - then @Mesh-ach , can you take that function I built build_and_log_ranked_feature_table, and separate it so we have:
> 
> 1. One function that builds the 3-column table (feature name, data type, and average SHAP magnitude). This is called at training and inference time.
> 2. A separate function (only used at training) that logs the 3-column table as an artifact. It needs to be named in the same way it is now: `selected_features/ranked_selected_features.csv`. Otherwise model card won't be able to retrieve it.
> 3. Then the third function uses that 3-column table then adds the two other columns (short desc and long desc). This will be called at both training and inference. This 5 column table will be the FE table at training and inference.
> 
> Let me know if this implementation works for you or if you have a more DRY approach, feel free to do that. @kaylawilding to confirm above looks right.

yeah got it, I'll take this up. Thanks Vish

